PR #88, opened 2019-04-09
add unit test for gring

aloncn:


johngcn:
@aloncn 覆盖率还差一点点，加油
image

PR #243, opened 2019-07-11
function BuildParams add 'UnUrlEncode' param, default false.

iReadX:
This function was originally forced to Url encoding, sometimes not needed; now add a parameter to enhance the flexibility of the function. When the new parameter has no value, the function is the same as before.


johngcn:
你这代码写得6

PR #311, opened 2019-09-04
Slightly improved usability of the module

LetMyPplGo:
1. added default port=25 for Address
2. added simple check that To contains '@'
3. made minor code makeups


net/gsmtp/gsmtp.go
```diff
@@ -38,35 +38,36 @@ func New(address, username, password string) *SMTP {
 // and then sends an email from address from, to addresses to, with
 // message msg.
 func (s *SMTP) SendMail(from, tos, subject, body string, contentType ...string) error {
-	if s.Address == "" {
-		return fmt.Errorf("address is necessary")
-	}
```
LetMyPplGo: This check is done later together with 'is port specified in the address?'

net/gsmtp/gsmtp.go
```diff
@@ -38,35 +38,36 @@ func New(address, username, password string) *SMTP {
...
-	for i := 0; i < count; i++ {
-		if arr[i] == "" {
-			continue
+	for _, to := range arr {
+		// TODO: replace with regex
+		if strings.Containts(to, "@") {
+			tosArr = append(tosArr, to)
 		}
-		safeArr = append(safeArr, arr[i])
 	}
```
LetMyPplGo: using range instead of for i:=0 which is more Go style
also renamed safeArr into tosArr as it will be used later

PR #312, opened 2019-09-04
Slightly improved usability of the module

LetMyPplGo:
1. added default port=25
2. added unit test for address
3. added simple Email address check
4. added example


johngcn:
@LetMyPplGo 
Thanks for your contribution, but the CI went failed, please have a check: https://travis-ci.org/gogf/gf/jobs/580949999

go.sum
```diff
@@ -2,6 +2,7 @@ github.com/AndreasBriese/bbloom v0.0.0-20190306092124-e2d15f34fcf9 h1:HD8gA2tkBy
 github.com/AndreasBriese/bbloom v0.0.0-20190306092124-e2d15f34fcf9/go.mod h1:bOvUY6CB00SOBii9/FifXqc0awNKxLFCL/+pkDPuyl8=
 github.com/BurntSushi/toml v0.3.1 h1:WXkYYl6Yr3qBf1K79EBnL4mak0OimBfB0XUf9Vl28OQ=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
+github.com/LetMyPplGo/gf v1.8.3 h1:hwVRTxfs1rWYOqCP6JbIZTf/wbM1SekAQ/nTGs+ELTA=
```
johngcn: Do not add this, you may consider: github.com/gogf/gf

LetMyPplGo: @johngcn, that was my first PR to github :)
Fixed the problems, should be fine now. Please see PR #316

LetMyPplGo: btw, it looks like I can't add the fix in the code and the unit test for that fix in the same PR. The new test will fail as it tests the old code - that's what happened here. How can I make the test check the new code?

PR #346, opened 2019-09-26
优化函数grand.Str

danvinhe:
The 'letters' already contain numbers, so just grab characters randomly from this.


johngcn:
@danvinhe Nice improvement!

PR #478, opened 2020-01-16
logo suggest

antoniasymeonidou:


gqcn:
@antoniasymeonidou Thank you for your contribution, but this logo looks a little bit normal, and it should be cute and impressive. It's suggested a logo a little like the one of Go, which is a mole. It does not matter whether mention the word `GoFrame` or not.

PR #601, opened 2020-03-31
Get ghttp a chance to add custom host.

bdliyq:
Get ghttp client a chance to add custom host in header. Golang can only set host with Request while it seems that there is not a possible way to add a custom host to a http request.

### DIFF
--- a/net/ghttp/ghttp_client_request.go
+++ b/net/ghttp/ghttp_client_request.go
@@ -186,12 +186,19 @@ func (c *Client) DoRequest(method, url string, data ...interface{}) (resp *Clien
 	// Custom header.
 	if len(c.header) > 0 {
 		for k, v := range c.header {
-			// Set host by req.Host.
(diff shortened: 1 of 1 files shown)


net/ghttp/ghttp_client_request.go
```diff
@@ -186,6 +186,10 @@ func (c *Client) DoRequest(method, url string, data ...interface{}) (resp *Clien
 	// Custom header.
 	if len(c.header) > 0 {
 		for k, v := range c.header {
+			// Set host by req.Host.
+			if strings.ToLower(k) == "host" {
+				req.Host = v
+			}
 			req.Header.Set(k, v)
 		}
```
gqcn: 这里遍历检查的效率不高，建议可以这样：
```go
if host := req.Header.Get("Host"); host != "" {
    req.Host = host
}
```
注意这里把注释写上。

PR #757, opened 2020-06-22
Fix typo in git ignore

leoleoasd:


leoleoasd:
Also, added newline at the end of this file, according to [the POSIX standard.](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206)

PR #978, opened 2020-11-05
Update gtimer_timer.go

lutherlau:
bugfix: avoid overflow


os/gtimer/gtimer_timer.go
```diff
@@ -233,7 +233,7 @@ func (t *Timer) binSearchIndex(n int64) (index int, result int) {
 	mid := 0
 	cmp := -2
 	for min <= max {
-		mid = int((min + max) / 2)
+		mid = min + int((max-min)/2)
```
gqcn, line 236: I do not see any bug here, would you please have a little bit explanation about this change?

gqcn (changes requested):

PR #979, opened 2020-11-05
Update garray_sorted_any.go

lutherlau:
bugfix: avoid overflow


container/garray/garray_sorted_any.go
```diff
@@ -446,7 +446,7 @@ func (a *SortedArray) binSearch(value interface{}, lock bool) (index int, result
 	mid := 0
 	cmp := -2
 	for min <= max {
-		mid = (min + max) / 2
+		mid = min + int((max-min)/2)
```
gqcn, line 449: @lutherlau Thanks for your PR, but I do not see any bug here, would you please have a little bit explanation about this change?

gqcn (changes requested):

PR #980, opened 2020-11-05
Update garray_sorted_str.go

lutherlau:
bugfix: avoid overflow

### DIFF
--- a/container/garray/garray_sorted_str.go
+++ b/container/garray/garray_sorted_str.go
@@ -445,7 +445,7 @@ func (a *SortedStrArray) binSearch(value string, lock bool) (index int, result i
 	mid := 0
 	cmp := -2
 	for min <= max {
-		mid = (min + max) / 2
+		mid = min + int((max-min)/2)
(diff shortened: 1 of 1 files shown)


container/garray/garray_sorted_str.go
```diff
@@ -445,7 +445,7 @@ func (a *SortedStrArray) binSearch(value string, lock bool) (index int, result i
 	mid := 0
 	cmp := -2
 	for min <= max {
-		mid = (min + max) / 2
+		mid = min + int((max-min)/2)
```
gqcn, line 448: @lutherlau Thanks for your PR, but I do not see any bug here, would you please have a little bit explanation about this change?

gqcn (changes requested):

container/garray/garray_sorted_str.go
lutherlau: yeah, reference this [essay](https://ai.googleblog.com/2006/06/extra-extra-read-all-about-it-nearly.html)

PR #983, opened 2020-11-06
Update gjson_api_new_load.go to make gjson work when sometimes the xml end with space or `\n`

coolhihi:
Make it work when the xml content has ` `,`\n`,`\r` or `\t` in the end. 

The json content can work well when the content not end with `}`, I think xml should not be too strict.

### DIFF
--- a/encoding/gjson/gjson_api_new_load.go
+++ b/encoding/gjson/gjson_api_new_load.go
@@ -228,7 +228,7 @@ func IsValidDataType(dataType string) bool {
 func checkDataType(content []byte) string {
 	if json.Valid(content) {
 		return "json"
-	} else if gregex.IsMatch(`^<.+>[\S\s]+<.+>$`, content) {
+	} else if gregex.IsMatch(`^<.+>[\S\s]+<.+>\s*$`, content) {
 		return "xml"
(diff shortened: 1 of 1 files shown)


encoding/gjson/gjson_api_new_load.go
```diff
@@ -228,7 +228,7 @@ func IsValidDataType(dataType string) bool {
 func checkDataType(content []byte) string {
 	if json.Valid(content) {
 		return "json"
-	} else if gregex.IsMatch(`^<.+>[\S\s]+<.+>$`, content) {
+	} else if gregex.IsMatch(`^<.+>[\S\s]+<.+>[ \n\r\t]*$`, content) {
```
gqcn, line 231: @coolhihi Maybe it could use `^<.+>[\S\s]+<.+>[\s]*$` instead better here?

gqcn (changes requested):

encoding/gjson/gjson_api_new_load.go
coolhihi: you're right. `[\s]*`

coolhihi: `\s*`

PR #987, opened 2020-11-09
update glog , support custom extName

zcool321:
https://github.com/gogf/gf/issues/984


gqcn:
@zcool321 It's already done.

PR #1033, opened 2020-12-02
support query string without value like ?abc&def=1

forrestjgq:
support query string without value like ?abc&def=1 in which abc is a  parameter without value


gqcn:
@forrestjgq Thanks for your PR. But the CI failed, please have a check the reason. Besides, would you please add some unit testing cases for your modification?

forrestjgq:
The failed CI case has conflicts with my fixing. Should I modify test case?

gqcn:
> The failed CI case has conflicts with my fixing. Should I modify test case?

Sure, just go ahead, and add your own testing case along with your PR.

gqcn:
Closed for no activity for too long.

PR #1058, opened 2020-12-12
improve function empty

viken2:


internal/empty/empty.go
```diff
@@ -104,13 +104,40 @@ func IsEmpty(value interface{}) bool {
 		} else {
 			rv = reflect.ValueOf(value)
 		}
+
 		switch rv.Kind() {
+		case reflect.Bool:
```
gqcn, line 109: @viken2 These checks are already handled in previous type assertions, why did you add the same checks using reflection here?
image

gqcn (changes requested):

internal/empty/empty.go
viken2: 目前的功能是无法支持自定义类型的判断，比如 type TestInt int，类似这种，var a TestInt = 0 就无法识别

gqcn: OK，I get it.

PR #1132, opened 2021-01-19
Support immutable collections.

syhily:
Add the immutable ability for `garray`, `gmap`, `gset`.

### DIFF
--- a/container/garray/garray_immutable_any.go
+++ b/container/garray/garray_immutable_any.go
@@ -1,0 +1,111 @@
+// Copyright GoFrame Author(https://goframe.org). All Rights Reserved.
+//
+// This Source Code Form is subject to the terms of the MIT License.
+// If a copy of the MIT was not distributed with this file,
+// You can obtain one at https://github.com/gogf/gf.
+
+package garray
+
+// NewImmutableArray would create a immutable array from go slice.
+// We recommend using the Array.Immutable() instead of this for better customization.
+func NewImmutableArray(array ...interface{}) ImmutableArray {
+	return NewArrayFromCopy(array, false)
+}
+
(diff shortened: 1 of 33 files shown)


gqcn:
@syhily I think it'd better using interface instead of defining new struct implementing immutable feature.

syhily:
Using immutable interface means that they share same struct, but this don't make the instance immutable. Case we can force cast the immutable instance to its real struct. Wrapping the mutable instance to a immutable would be better on this scenario.

gqcn:
> Using immutable interface means that they share the same struct, but this don't make the instance immutable. Case we can force cast the immutable instance to its real struct. Wrapping the mutable instance into an immutable struct would be better on this scenario.

There're so many repeated codes between the containers and their immutable feature version. Using interface design implementing immutable feature makes few and easy-maintaining codes for the containers. The immutable  interface can export some read-only method in its definition to make its container immutable. I think you may create a new container struct using `Clone` function and return it as an immutable interface to reduce the repeated and unnecessary codes.

syhily:
> > Using immutable interface means that they share the same struct, but this don't make the instance immutable. Case we can force cast the immutable instance to its real struct. Wrapping the mutable instance into an immutable struct would be better on this scenario.
> 
> There're so many repeated codes between the containers and their immutable feature versions. Using interface design implementing immutable feature makes few and easy-maintaining codes for the containers. The immutable interface can export some read-only method in its definition to make its container immutable. I think you may create a new container struct using `Clone` function and return it as an immutable interface to reduce the repeated and unnecessary codes.

This could be the best solution. I will push the new code today.

gqcn:
@syhily After thinking of it many times, I feel that the implementation of this feature is not graceful, so I decided not to merge this feature into the main branch.

PR #1150, opened 2021-01-31
formatMonthDaySuffixMap() misjudged suffix

gouguoyin:
when day is 21，abbreviated as 21st，suffix is st ,is not th
when day is 22，abbreviated as 22nd，suffix is nd ,is not th
when day is 23，abbreviated as 23rd，suffix is rd ,is not th
when day is 31，abbreviated as 31st，suffix is st ,is not th


gouguoyin:
retry

PR #1315, opened 2021-06-28
add response WriteProtoBuf func

biu7:


gqcn:
@biu7 Would you please give some brief why it is necessary adding protobuf marshaling response feature for ghttp.Server?

biu7:
> @biu7 Would you please give some brief why it is necessary adding protobuf marshaling response feature for ghttp.Server?

we use gf to transferProtobuf  instead of json, we have to write some internal function to achieve that, and the code is not elegant. So we really want such a feature.

gqcn:
@biu7 It's very appreciate of your PR, but we cannot accept protobuf into the basic workframe as the it is not the common and necessary usage for most of the projects. Your can use inheriting feature of Golang struct or wrap a utility function to achieve your protobuf response purpose.

PR #1400, opened 2021-09-09
Fix bugs of gring

danvinhe:
1. Unlink should return a new  pointer of object
2. Len return bad count


gqcn:
@danvinhe Thanks for your PR. But there're some deep details I should fix, so I fix this and close this PR.

danvinhe:
> @danvinhe Thanks for your PR. But there're some deep details I should fix, so I fix this and close this PR.

ok

PR #1407, opened 2021-09-17
Add STRAIGHT_JOIN support

isword123:
Refer: https://stackoverflow.com/questions/512294/when-to-use-straight-join-with-mysql

### DIFF
--- a/database/gdb/gdb_model_join.go
+++ b/database/gdb/gdb_model_join.go
@@ -93,3 +93,38 @@ func (m *Model) doJoin(operator string, table ...string) *Model {
 	}
 	return model
 }
+
+// StraightJoin use STRAIGHT_JOIN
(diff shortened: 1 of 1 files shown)


database/gdb/gdb_model_join.go
```diff
@@ -93,3 +93,38 @@ func (m *Model) doJoin(operator string, table ...string) *Model {
 	}
 	return model
 }
+
+// StraightJoin use STRAIGHT_JOIN
+// Refer: https://stackoverflow.com/questions/512294/when-to-use-straight-join-with-mysql
+func (m *Model) StraightJoin(table ...string) *Model {
```
gqcn, line 99: Please implement the logic in function `doJoin`, add detailed comment and corresponding unit testing case for your feature.

gqcn (changes requested):

PR #1487, opened 2021-11-18
add ci notice for go fmt

stardemo:


stardemo:
ci use sh ,commond not effect

PR #1535, opened 2021-12-28
support manipulate pg.Array type data

starccck:
Implementing ConvertDataForRecord in DriverPgsql to support pg.Array


gqcn:
@alaywn 看起来不错，建议确认一下

PR #1546, opened 2022-01-04
Recurse map fields inside structs

eh-steve:


gqcn:
@eh-steve Please add according UT cases for your updates.

eh-steve:
> @eh-steve Please add according UT cases for your updates.

I've added some - the main problem case was for handling nested maps, and the specific example I was hitting was when converting from [a structure unmarshalled from YAML with merged anchors being converted to JSON](https://github.com/go-yaml/yaml/issues/139#issuecomment-566073291)

Issue #1655, opened 2022-03-09
Get scheme of the current request URL


houseme:

houseme:
https://github.com/golang/go/issues/28940

houseme:
https://stackoverflow.com/questions/40826664/get-scheme-of-the-current-request-url

https://github.com/oauth2-proxy/oauth2-proxy/pull/285

https://github.com/traefik/traefik/pull/6226

houseme:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto

https://www.nginx.com/resources/wiki/start/topics/examples/forwarded/

gqcn:
https://github.com/gogf/gf/pull/1657

PR #1657, opened 2022-03-09
fix: server access logs contain the protocol used between the server …

houseme:
…and the load balancer, but not the protocol used between the client and the load balancer

https://github.com/gogf/gf/issues/1655

### DIFF
--- a/net/ghttp/ghttp_request.go
+++ b/net/ghttp/ghttp_request.go
@@ -224,8 +224,11 @@ func (r *Request) GetRemoteIp() string {
 
 // GetUrl returns current URL of this request.
 func (r *Request) GetUrl() string {
-	scheme := "http"
-	proto := r.Header.Get("X-Forwarded-Proto")
+	var (
+		scheme = "http"
+		proto  = r.Header.Get("X-Forwarded-Proto")
+	)
+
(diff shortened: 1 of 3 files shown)


net/ghttp/ghttp_request.go
```diff
@@ -224,8 +224,12 @@ func (r *Request) GetRemoteIp() string {
...
 // GetUrl returns current URL of this request.
 func (r *Request) GetUrl() string {
-	scheme := "http"
-	if r.TLS != nil {
+	var (
+		scheme = "http"
+		proto  = r.Header.Get("X-Forwarded-Proto")
+	)
+
+	if r.TLS != nil || (proto != "" && strings.ToLower(proto) == "https") {
```
gqcn, line 232: `if r.TLS != nil ||  gstr.Equal(proto, "https") {`

gqcn: You should also update here: https://github.com/gogf/gf/blob/8f326dcac5548d85d465a135b6fa74af70a2be1c/net/ghttp/ghttp_server_log.go#L22

gqcn (changes requested):

net/ghttp/ghttp_request.go
houseme: Handled

PR #1659, opened 2022-03-10
[fix bug] the default value of r.get is invalid

arieslee:
[fix bug] the default value of r.get is invalid
重跑ci

### DIFF
--- a/net/ghttp/ghttp_request_param_request.go
+++ b/net/ghttp/ghttp_request_param_request.go
@@ -41,7 +41,7 @@ func (r *Request) GetRequest(key string, def ...interface{}) *gvar.Var {
 	if value.IsNil() {
 		value = r.GetRouter(key)
 	}
-	if value.IsNil() {
+	if !value.IsNil() {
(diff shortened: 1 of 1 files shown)


net/ghttp/ghttp_request_param_request.go
```diff
@@ -24,24 +24,24 @@ import (
...
-	if value == nil {
+	if value.IsNil() {
 		value = r.GetQuery(key)
 	}
-	if value == nil {
+	if value.IsNil() {
 		value = r.GetRouter(key)
 	}
-	if value != nil {
+	if value.IsNil() {
```
gqcn, line 44: `if !value.IsNil() {`

gqcn (changes requested):

net/ghttp/ghttp_request_param_request.go
arieslee: 哈哈哈哈，手误，刚才我自己也发现了

PR #1664, opened 2022-03-10
add try files config & handler

stardemo:
add support try_files fucn,to fit vue/react history router 
https://router.vuejs.org/zh/guide/essentials/history-mode.html#nginx


gqcn:
@stardemo The configuration `IndexFiles` works as `tryfiles` logic.

PR #1678, opened 2022-03-17
Add gmath

CrazyRocks:
frequently used math functions


gqcn:
@CrazyRocks Thanks for PR, but I see most of the content is alias of package `math/cmplx`. So I do not think it is necessary.

Issue #1714, opened 2022-03-27
timezone bug when persisting *gtime.Time to db


WesleyWu:
### 1. What version of `Go` and system type/arch are you using?

go1.17.6 darwin/amd64

### 2. What version of `GoFrame` are you using?

v2.0.4

### 3. Can this issue be re-produced with the latest release?
yes

### 4. What did you do?

1. database table
```
create table test_time (
  id int unsigned auto_increment,
  my_timestamp TIMESTAMP,
  my_time DATETIME,
  primary key(id)
);
```

2. write a simple service "TestTime" to save record to db

3. DSN
`mysql:user:pass@tcp(127.0.0.1:3306)/db?charset=utf8mb4&parseTime=true&loc=Local`

4. Use *time.Time for column type, and write a test as below:
```
func Test_insertDate(t *testing.T) {
	var time *time.Time
	time = &gtime.NewFromStr("2022-03-27T01:03:04Z").Time
	fmt.Println(time.String())
	service.TestTime.Add(gctx.New(), &model2.TestTimeAddReq{
		MyTimestamp: time,
		MyTime:      time,
	})

	time = &gtime.NewFromStr("2022-03-27 01:03:04").Time
	fmt.Println(time.String())
	service.TestTime.Add(gctx.New(), &model2.TestTimeAddReq{
		MyTimestamp: time,
		MyTime:      time,
	})

	time = &gtime.NewFromStr("2022-03-27T01:03:04+08:00").Time
	fmt.Println(time.String())
	service.TestTime.Add(gctx.New(), &model2.TestTimeAddReq{
		MyTimestamp: time,
		MyTime:      time,
	})

	time = &gtime.NewFromStr("2022-03-27T01:03:04+01:00").Time
	fmt.Println(time.String())
	service.TestTime.Add(gctx.New(), &model2.TestTimeAddReq{
		MyTimestamp: time,
		MyTime:      time,
	})

	time = &gtime.NewFromStr("2022-03-27T01:03:04+00:00").Time
	fmt.Println(time.String())
	service.TestTime.Add(gctx.New(), &model2.TestTimeAddReq{
		MyTimestamp: time,
		MyTime:      time,
	})
}
```

console output:
```
2022-03-27 01:03:04 +0000 UTC
2022-03-27 01:03:04 +0800 CST
2022-03-27 01:03:04 +0800 CST
2022-03-27 00:03:04 +0000 UTC
2022-03-27 01:03:04 +0000 UTC
```

database records:
```
[...]

PR #1718, opened 2022-03-28
fix timezone bug when persisting *gtime.Time to db #1714

WesleyWu:
`gtime_time_wrapper.go`
should
`return t.Local().Format("2006-01-02 15:04:05.999999999")`
instead of
`return t.Format("2006-01-02 15:04:05")`


gqcn:
@WesleyWu I think you may check the point of `gtime.Time` variable converting  before the variable committed to underlying driver in package gdb, not modifying the package `gtime`.

Issue #1727, opened 2022-04-03
Bug using *time.Time go type when retrieving db timestamp column value


WesleyWu:
### 1. What version of `Go` and system type/arch are you using?

go1.17.6 darwin/amd64

### 2. What version of `GoFrame` are you using?

v2.0.4 and above

### 3. Can this issue be re-produced with the latest release?
yes

### 4. What did you do?

My test case:

```
// create table test_time (
// id int unsigned auto_increment,
// my_gtime TIMESTAMP,
// my_time TIMESTAMP,
// my_time_ptr TIMESTAMP,
// primary key(id)
//);
func Test_selectDate(t *testing.T) {
	var (
		ctx     = gctx.New()
		myGtime = gtime.NewFromStr("2022-03-27 01:03:04")        // 2022-03-27 01:03:04 本地时间，目前是 CST +08:00
		myTime  = time.Date(2022, 3, 27, 1, 3, 4, 0, time.Local) // 2022-03-27 01:03:04 +0800 CST
	)

	g.Log().Debugf(ctx, "gtime=%s time=%s", myGtime.String(), myTime.String())
	_, _ = g.DB().Insert(ctx, "test_time", g.Map{
		"id":          10088,
		"my_gtime":    myGtime,
		"my_time":     myTime,
		"my_time_ptr": &myTime,
	})

	type TestTime struct {
		MyGtime   *gtime.Time
		MyTime    time.Time
		MyTimePtr *time.Time
	}

	var testRes = TestTime{}

	_ = g.DB().Model("test_time").WherePri(10088).Scan(&testRes)
	assert.Equal(t, myGtime, testRes.MyGtime)  // pass
	assert.Equal(t, myTime, testRes.MyTime)      // pass
	assert.Equal(t, &myTime, testRes.MyTimePtr)  // will fail
}

```

### 5. What did you expect to see?
test pass

### 6. What did you see instead?
failed with message

```
=== RUN   Test_selectDate
2022-04-03 14:36:38.456 [DEBU] {4833c3e0794fe21650260717942a6c21} gtime=2022-03-27 01:03:04 time=2022-03-27 01:03:04 +0800 CST 
2022-04-03 14:36:38.506 [DEBU] {4833c3e0794fe21650260717942a6c21} [ 48 ms] [default] [rows:4  ] SHOW FULL COLUMNS FROM `test_time` 
[...]

PR #1729, opened 2022-04-03
fix timezone bug when persisting *gtime.Time to db #1714

WesleyWu:


WesleyWu:
> @WesleyWu If you need auto time converting feature of underlying driver, please use `time.Time` instead of `gtime.Time`.
@gqcn 
If I use *time.Time for columns, I could not retrieve them right because issue #1727.
Not considering to use time.Time because I need to check if column value is nil.

Thanks.

WesleyWu:
> @WesleyWu If you need auto time converting feature of underlying driver, please use `time.Time` instead of `gtime.Time`.

@gqcn 
And I don't think I'm using any auto time conversion because my fix did not do any conversion, which should be expected behavior for db persisting. On the contrary, the original implementation used time conversion by calling gtime.Time.String().

WesleyWu:
the gtime.Time.String() method is flawed because it loses timezone info.

PR #1757, opened 2022-04-13
gvalid.aliasNameTagPriority consistent with gconv.StructTagPriority.

yuancjun:
This is a sample that causes an assertion error.

```go
package main

import (
	"github.com/gogf/gf/v2/os/gctx"
	"github.com/gogf/gf/v2/test/gtest"
)

func main() {
	type Req struct {
		Name string `json:"n" v:"required"`
	}

	ctx := gctx.New()
	req := Req{}
	data := g.MapStrStr{
		"n": "v1",
	}
	gtest.AssertNil(g.Validator().Data(&req).Assoc(data).Run(ctx))
}
```

### DIFF
--- a/util/gvalid/gvalid.go
+++ b/util/gvalid/gvalid.go
@@ -183,8 +183,8 @@ var (
 		"no":    {},
 	}
 
-	structTagPriority    = []string{"gvalid", "valid", "v"} // structTagPriority specifies the validation tag priority array.
-	aliasNameTagPriority = []string{"param", "params", "p"} // aliasNameTagPriority specifies the alias tag priority array.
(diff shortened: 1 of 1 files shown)


gqcn:
@yuancjun The validation component does not consider `json` tag as its alias name, but you can use validation rule tag `gvalid tag` to do so. Reference: https://goframe.org/pages/viewpage.action?pageId=7302150
Here's a example:
```go
package main

import (
	"github.com/gogf/gf/v2/os/gctx"
	"github.com/gogf/gf/v2/test/gtest"
)

func main() {
	type Req struct {
		Name string `v:"n#required"`
	}

	ctx := gctx.New()
	req := Req{}
	data := g.MapStrStr{
		"n": "v1",
	}
	gtest.AssertNil(g.Validator().Data(&req).Assoc(data).Run(ctx))
}
```

PR #1762, opened 2022-04-14
upgrade ClickHouse dependencies to V2

DGuang21:


contrib/drivers/clickhouse/clickhouse.go
```diff
@@ -37,7 +38,9 @@ var (
 	errUnsupportedReplace      = errors.New("unsupported method:Replace")
 	errUnsupportedBegin        = errors.New("unsupported method:Begin")
 	errUnsupportedTransaction  = errors.New("unsupported method:Transaction")
-	errSQLNull                 = errors.New("SQL cannot be null")
+	errUpdateNotCondition      = errors.New("there should be WHERE condition statement for UPDATE operation")
+	errDeleteNotCondition      = errors.New("there should be WHERE condition statement for DELETE operation")
+	errUpdateNotAssignment     = errors.New("there should be WHERE condition statement for Assignment operation")
```
gqcn, line 43: These errors already handled by package gdb, it is not necessary doing the same checks.

contrib/drivers/clickhouse/clickhouse_test.go
```diff
@@ -3,6 +3,7 @@ package clickhouse
 import (
 	"context"
 	"fmt"
+	"github.com/google/uuid"
```
gqcn, line 6: It is possible using package `guid` instead of `uuid`?

gqcn (changes requested):

PR #1768, opened 2022-04-18
.Properties Setting file not support

happyinsect:
.Properties Setting file not support,  which is widely used in Java project.


gqcn:
@happyinsect Please fix conflicts.

gqcn:
Duplicated with https://github.com/gogf/gf/pull/1806

PR #1775, opened 2022-04-19
no longer use unsafe package for string/bytes conversion

DGuang21:


.github/workflows/go.yml
```diff
@@ -3,10 +3,11 @@ name: GoFrame CI
...
   push:
     branches:
-      - master
-      - develop
+      - 'master'
+      - 'develop'
+      - 'fix-unsafe'
   pull_request:
-    branches: [master, develop]
+    branches: ['master', 'develop','fix-unsafe']
```
gqcn, line 10: `'fix-unsafe'` -> ``fix/*``

gqcn (changes requested):

PR #1891, opened 2022-06-03
Keep DaoPath and DoPath the same as the document.

rocna:
Keep DaoPath and DoPath the same as the document.


gqcn:
@rocna It is changed in version from `v2.1`

rocna:
NIce

PR #1944, opened 2022-06-27
add cross building support for sqlite in command gen dao

gqcn:
add cross building support for sqlite in command gen dao


hailaz (approved):
没问题

PR #1950, opened 2022-06-28
Dev gen service

L-fushen:
gf gen service supports the generation of service files in the specified naming format


L-fushen:
gf gen service supports the generation of service files in the specified naming format

PR #1957, opened 2022-06-29
convert gconv.String([]rune) to string, like []byte

dianbanjiu:


gqcn:
@dianbanjiu CI failed.

gqcn:
No activity for more than one month.

PR #2009, opened 2022-07-15
fix goai_path.go requestBodyContent.Schema nil panic

ar026:


gqcn:
@ar026 Please `gofmt` your  changes before pr.

PR #2030, opened 2022-07-26
add WritePureJson for ghttp.Response

felix-Lu:


hailaz (approved):

gqcn:
@felix-Lu Thanks for your PR. You can use `Response.Write` along with `Response.Header().Set("Content-Type", contentTypeJson)` to do so, instead of adding new function, which is unnecessary and making it more complicated for JSON response.

PR #2032, opened 2022-07-26
add MiddlewareJsonBody, improve error response handling for package ghttp

gqcn:


net/ghttp/ghttp.go
```diff
@@ -12,6 +12,8 @@ import (
 	"reflect"
 	"time"
 
+	"github.com/gogf/gf/v2/errors/gcode"
+	"github.com/gogf/gf/v2/errors/gerror"
```
houseme, line 16: gofmt

gqcn: It's already `gofmt -w`, it or else could not go through CU procedure.

houseme (approved):
LGTM

PR #2053, opened 2022-08-06
add function Warn/Warnf for glog

LingCoder:


gqcn:
@LingCoder The logging method names `Warning/Warningf` are more frequently used than `Warn/Warnf`. Please refer to PHP Laravel:  https://laravel.com/docs/9.x/logging , and Python: https://docs.python.org/3.8/howto/logging.html

PR #2064, opened 2022-08-12
fixed inconsistent results when converting float64(NaN) to int/uint on multiple platforms

DGuang21:
Mainly include the following changes
- When using gconv.Int64/gconv.Uint64 , if the result is NaN, it will become 0
- When using gconv.Float64 / gconv.Float32 , if the result is NaN, it will not be modified

Fix the change only occurs when floating point to uint/int

### DIFF
--- a/util/gconv/gconv_int.go
+++ b/util/gconv/gconv_int.go
@@ -7,6 +7,7 @@
 package gconv
 
 import (
+	"math"
 	"strconv"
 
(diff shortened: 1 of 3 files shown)


gqcn:
@DGuang21 这个`NaN`在Golang标准库的整型转换中是一个特殊的字符串，我个人觉得改变这种转换行为是否得当？

gqcn (approved):

PR #2113, opened 2022-09-08
Allow manual forced update of field information

rzhghost:
example:
`value, err := db.Model("user").Fields("passport").Where("id=3").Value()`
`......`
`if _, err := db.Exec(ctx, "ALTER TABLE user ADD COLUMN newField1 varchar(50) NULL")
       err != nil {
		gtest.Error(err)
	}`
`......`
`_, err := g.DB().TableFields(context.WithValue(ctx, "refresh_fields", true), "user")`
`    if err != nil {`
`	    return err`
`    }`


gqcn:
@rzhghost CI failed.

rzhghost:
fixed

gqcn:
@rzhghost Thanks for PR, but it is duplicated with https://github.com/gogf/gf/pull/2128

rzhghost:
@gqcn  It is better to specify a table name to clear the cache to avoid frequent execution of "SHOW FULL COLUMNS"

PR #2133, opened 2022-09-20
refract builtin rules management mechanism, add `eq/not-eq/gt/gte/lt/lte/before/before-equal/after/after-equal/array/not-regex` rules for for package `gvalid`

gqcn:
1. refract builtin rules management mechanism for package `gvalid`
2. add `eq/not-eq/gt/gte/lt/lte/before/before-equal/after/after-equal/array/not-regex` rules for for package `gvalid`


houseme (approved):
LGTM

PR #2144, opened 2022-09-24
improve `contrib/drivers/dm`

gqcn:


zhonghuaxunGM (approved):
Adjusting the schema is required and the upper layer does not need to specify the schema

PR #2205, opened 2022-10-14
fix gen dao oracle entity column datatype number bug

zaier277:
bug:When oracle column datatype is number , gen dao entity column type is string.

fix: when  (DATA_TYPE='NUMBER' AND DATA_SCALE=0)  set datatype is int .   when  (DATA_TYPE='NUMBER' AND DATA_SCALE>0)  set datatype is float


gqcn:
@zaier277 It's appreciated for your PR! The UT case failed in CI.

PR #2212, opened 2022-10-15
Feat / check error when creating new db(＃2204)

### LINKED ISSUE #2204: error from gdb.New()  always nil
```go
node := gdb.ConfigNode{
		  Host:  "127.0.0.1",
		  Port:  "9000",
		  User:  "default",
		  Name:  "default",
		  Type:  "clickhouse",
		  Debug: false,
	}
db, err = gdb.New(node)
if err ! = nil{
     
}


// New creates and returns a database object for clickhouse.
// It implements the interface of gdb.Driver for extra database driver installation.
func (d *Driver) New(core *gdb.Core, node *gdb.ConfigNode) (gdb.DB, error) {
	return &Driver{
		Core: core,
	}, nil
}
```
### 5. What did you expect to see?

if fail to connect database, I wish err ! = nil
 
### 6. What did you see instead?
The returned err is nil regardless of whether the connection is successful or not
（err 始终为nil ）

CCpro10:
resolves #2204


CCpro10:
cc@gqcn

PR #2213, opened 2022-10-15
Parse database empty json field to empty object( fix issue 2105 )

zhonghuaxunGM:
My change is just to satisfy the orm situation in issue, but I'm not sure if it will cause secondary problems in other uses. I have doubts about my changes, please help me to judge and solve.


gqcn:
@zhonghuaxunGM It's appreciated for your PR! I've committed some changes directly to your branch to fix issue #2105.

houseme (approved):

PR #2215, opened 2022-10-17
Whether the logic directory configuration can be set to support multiple directories like dao

mijjjj:
I have a problem now. I want to define multiple app directories, but the logic supported by default has only one path. Can the client command support multiple paths like dao?


houseme:
PR 修改的内容是什么呢？

PR #2219, opened 2022-10-17
fix gen dao oracle entity bug

zaier277:
When Oracle  Column Datatype is NUMBER check DATA_SCALE
DATA_SCALE==0  use INT
DATA_SCALE>0    use float64


gqcn:
@zaier277 CI failed.

gqcn (approved):

PR #2220, opened 2022-10-18
attach and export real handler for request

SUDOCS:
record user defined handler to request object before executing middlewares, which makes it possible to do something special according to the final target router in handler.

#2217


net/ghttp/ghttp.go
```diff
@@ -91,8 +91,8 @@ type (
 		Source     string          // Registering source file `path:line`.
 	}
 
-	// handlerParsedItem is the item parsed from URL.Path.
-	handlerParsedItem struct {
+	// HandlerParsedItem is the item parsed from URL.Path.
+	HandlerParsedItem struct {
```
gqcn, line 95: 如果这个类型要对外开放，那么名称可以调整为`HandlerItemParsed`，与`HandlerItem`对应。

gqcn (changes requested):

net/ghttp/ghttp.go
SUDOCS: Thank you. I have renamed it to `HandlerParsedItem`.

gqcn (approved):

PR #2264, opened 2022-11-04
import clickhouse driver for client tool

wuyongshang:


DGuang21:
This feature has been completed in the PR (https://github.com/gogf/gf/pull/2214), import the mod is not good to generate the model/dao, need to change more than a little.

cmd/gf/internal/cmd/cmd_gen_dao.go
```diff
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	_ "github.com/gogf/gf/contrib/drivers/clickhouse/v2"
```
DGuang21, line 4: Try to generate code after introducing the clickhouse-go driver. An error may occur during direct generation.

DGuang21 (changes requested):

gqcn:
@wuyongshang @DGuang21 I think it should ignore the lower version (1.15/1.16/1.17) of go in CI.

cmd/gf/internal/cmd/cmd_gen_dao.go
wuyongshang: We used clickhouse for data storage in our new project. when I executed `gf gen dao` command, no new files were generated, as a new framework users of gf which was very confusing for me.

wuyongshang: If there are a lot of changes to solve this issue, please close this pull request

PR #2305, opened 2022-11-17
Update gutil_dump.go

brick-carrier-fan:
Solve the panic: reflect: call of reflect.Value.Type on zero Value [recovered] The reason for this is the circular reference itself

### DIFF
--- a/util/gutil/gutil_dump.go
+++ b/util/gutil/gutil_dump.go
@@ -110,9 +110,12 @@ func doDump(value interface{}, indent string, buffer *bytes.Buffer, option doDum
 	}
 	var (
 		reflectKind     = reflectValue.Kind()
-		reflectTypeName = reflectValue.Type().String()
+		reflectTypeName = "invalid"
 		newIndent       = indent + dumpIndent
 	)
+	if reflectValue.IsValid() {
+		reflectTypeName = reflectValue.Type().String()
+	}
 	reflectTypeName = strings.ReplaceAll(reflectTypeName, `[]uint8`, `[]byte`)
 	if !option.WithType {
 		reflectTypeName = ""


brick-carrier-fan:
it likes
type Abc struct {
		ab int
		//bc BC
		cd *Abc
	}
	
	abc := Abc{ab: 3}
	abc.cd = &abc
	g.Dump(abc)
output:
panic: reflect: call of reflect.Value.Type on zero Value

util/gutil/gutil_dump.go
```diff
@@ -110,9 +110,12 @@ func doDump(value interface{}, indent string, buffer *bytes.Buffer, option doDum
 	}
 	var (
 		reflectKind     = reflectValue.Kind()
-		reflectTypeName = reflectValue.Type().String()
+		reflectTypeName = "invalid"
```
gqcn, line 113: It may check the reflectValue at line 107:
```go
// Double check nil value.
if value == nil || reflectValue.IsNil() || !reflectValue.IsValid() {
	buffer.WriteString(`<nil>`)
	return
}
```

gqcn (changes requested):

gqcn:
@brick-carrier-fan Please add UT case for this issue, too.

util/gutil/gutil_dump.go
brick-carrier-fan: I've tried your way of checking the reflectValue, and it brings the same mistake.
Not available to check reflectValue with IsNil().
The method I submitted is available.

brick-carrier-fan: Yes,buffer will go to get infinitely bigger,then get OOM panic

gqcn: > Yes,buffer will go to get infinitely bigger,then get OOM panic
 
Any question carrying on this PR?

gqcn (changes requested):

gqcn:
@brick-carrier-fan I fix this in https://github.com/gogf/gf/pull/2367

Issue #2306, opened 2022-11-17
BuildParams() check urlEncode


primexiao:
https://github.com/gogf/gf/blob/182a3930500fab596028720be529f5eff74bde27/internal/httputil/httputils.go#L62

Please confirm that when value is not a fileupload, but length < 6 ... 
Event if urlEncode is true, this value will be ignore

Is it a issue?

PR #2308, opened 2022-11-17
fix BuildParams with urlEncode when len(v) <= 6

primexiao:
#2306 

Some typical cases:

```
func TestBuildParams(t *testing.T) {
        // normal && special cases
	params := map[string]string{
		"code":  "x",
		"code1": "x&a=1",   // for fix
		"code2": "x&a=111",
		"id":    "1+- ",  // for fix
		"f":     "1#a=+- ",
		"v":     "",
		"n":     "null",
	}

	res1 := ghttp.BuildParams(params)
	vs, err := url.ParseQuery(res1)
	assert.Truef(t, err == nil && len(params) == len(vs), "expected-len vs actual-len: [%d] vs [%d]", len(params), len(vs))
	for k := range vs {
		vv := vs.Get(k)
		_, ok := params[k]
		assert.Truef(t, ok && params[k] == vv, "[%s] expected vs actual: [%v] vs [%v]", k, params[k], vv)
	}
}

```


internal/httputil/httputils.go
```diff
@@ -59,7 +59,7 @@ func BuildParams(params interface{}, noUrlEncode ...bool) (encodedParamStr strin
 			encodedParamStr += "&"
 		}
 		s = gconv.String(v)
-		if urlEncode && len(s) > 6 && strings.Compare(s[0:6], fileUploadingKey) != 0 {
+		if urlEncode && (len(s) <= 6 || strings.Compare(s[0:6], fileUploadingKey) != 0) {
 			s = gurl.Encode(s)
```
gqcn, line 63: It may be:
```go
if urlEncode {
	if len(s) > 6 && strings.Compare(s[0:6], fileUploadingKey) == 0 {
		// No url encoding if uploading file.
	} else {
		s = gurl.Encode(s)
	}
}
```

gqcn (changes requested):

gqcn:
@primexiao Please add UT case for your changes.

gqcn:
@primexiao Any process?

internal/httputil/httputils.go
primexiao: 🆗  I just want to keep its original style in my first commit, but I agree with u more : )

internal/httputil/httputil_z_test.go
```diff
@@ -0,0 +1,36 @@
...
+// If a copy of the MIT was not distributed with this file,
+// You can obtain one at https://github.com/gogf/gf.
+
+package httputil
+
+import (
+	"net/url"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
gqcn, line 13: Please use `gtest` for simple testing cases, not import `testify` into goframe, referring to other ut cases for example.

Add the ut case to file `ghttp_z_unit_test.go`, not creating a new ut file.

gqcn (changes requested):

primexiao:
Something wrong with CI and I don't know why ... Seems that the details are about `gcache`

gqcn (approved):

PR #2402, opened 2023-01-17
gdb error should wrap original underlying database error like MySQLError

WesleyWu:
Sometimes we need to get underlying database error to check if it is a PKConflict or DataTooLong error(for MySQL).
GoFrame may better wrap the underlying error for further checking.

```
func DbErrorToRequestError(req interface{}, err error, db gdb.DB) (error, bool) {
	if err == nil {
		return nil, false
	}
	underlyingError := gerror.Unwrap(err)
	if underlyingError == nil {
		return nil, false
	}
	dbType := db.GetConfig().Type
	switch dbType {
	case "mysql":
		if driverError, ok := underlyingError.(*mysql.MySQLError); ok {
			switch driverError.Number {
			case 1062:
				return NewPkConflictErrorf(req, "%v: %v", "主键冲突", driverError.Error()), true
			case 1406:
				return NewDataTooLongErrorf(req, "%v: %v", "数据过长", driverError.Error()), true
			}
		}
	}
	return err, false
}
```


gqcn (approved):

database/gdb/gdb_core_underlying.go
```diff
@@ -278,8 +278,9 @@ func (c *Core) DoCommit(ctx context.Context, in DoCommitInput) (out DoCommitOutp
 		c.writeSqlToLogger(ctx, sqlObj)
 	}
 	if err != nil && err != sql.ErrNoRows {
-		err = gerror.NewCodef(
+		err = gerror.WrapCodef(
 			gcode.CodeDbOperationError,
+			err,
 			"%s, %s",
 			err.Error(),
```
gqcn, line 285: If it uses `Wrap`, it then should remove the duplicated error string from its error message.

gqcn (changes requested):

database/gdb/gdb_core_underlying.go
WesleyWu: Thanks. I'll make the modification.

WesleyWu (commented):
done.

database/gdb/gdb_core_underlying.go
```diff
@@ -278,10 +278,10 @@ func (c *Core) DoCommit(ctx context.Context, in DoCommitInput) (out DoCommitOutp
...
 	}
 	if err != nil && err != sql.ErrNoRows {
-		err = gerror.NewCodef(
+		err = gerror.WrapCodef(
 			gcode.CodeDbOperationError,
-			"%s, %s",
-			err.Error(),
+			err,
+			"%s",
 			FormatSqlWithArgs(in.Sql, in.Args),
```
gqcn, line 285: Is it better using 
`gerror.WrapCode(gcode.CodeDbOperationError, err, FormatSqlWithArgs(in.Sql, in.Args))`
rather than 
`gerror.WrapCodef(gcode.CodeDbOperationError, err, "%s", FormatSqlWithArgs(in.Sql, in.Args))`
?

gqcn (changes requested):

database/gdb/gdb_core_underlying.go
WesleyWu: sure

gqcn: @WesleyWu Should it use `gerror.WrapCode` here?

gqcn (changes requested):

gqcn (approved):

PR #2404, opened 2023-01-17
fix issue 2403

### LINKED ISSUE #2403: gdb使用结构查询，包含gjson.Json字段且为nil时，也会作为空字符串条件查询
gf v2.3.0

### 示例代码
```go
func main() {                                    
    type where struct {                          
        g.Meta `orm:"table:test_json, do:true"`
        ID   int                                 
        Data *gjson.Json                         
    }                                            
                                                 
    ctx := gctx.New()                            
                                                 
    query := where{                              
        ID: 1,                                   
    }                                            
                                                 
    g.DB().Model("test_json").Ctx(ctx).One(query)
}                                                
```

### 期望SQL
```sql
SELECT * FROM "test_json" WHERE "id"=1 LIMIT 1
```

### 实际执行SQL
```sql
SELECT * FROM "test_json" WHERE "id"=1 AND "data"='' LIMIT 1
```

qinyuguang:
fix #2403


database/gdb/gdb_func.go
```diff
@@ -210,7 +211,7 @@ func DataToMapDeep(value interface{}) map[string]interface{} {
 	m := gconv.Map(value, structTagPriority...)
 	for k, v := range m {
 		switch v.(type) {
-		case time.Time, *time.Time, gtime.Time, *gtime.Time:
+		case time.Time, *time.Time, gtime.Time, *gtime.Time, gjson.Json, *gjson.Json:
```
gqcn, line 214: Please add associated unit testing case for this case.

gqcn (changes requested):

database/gdb/gdb_func.go
qinyuguang: done

gqcn (approved):

PR #2415, opened 2023-02-04
Cache grouping optimization

z5830865:
Cache grouping optimization


z5830865 (commented):
Cache grouping optimization

gqcn:
@z5830865 Would you please give some description and business scenario about the `Group` feature?

z5830865:
> @z5830865 Would you please give some description and business scenario about the `Group` feature?

Cache grouping allows you to group the associated caches together, similar to the clear function. You can group all caches of a table into a group. Reading the source code, clear does not implement such a feature

database/gdb/gdb_model_cache.go
```diff
@@ -140,6 +172,16 @@ func (m *Model) saveSelectResultToCache(
 	if errCache := cacheObj.Set(ctx, cacheKey, cacheItem, m.cacheOption.Duration); errCache != nil {
 		intlog.Errorf(ctx, `%+v`, errCache)
 	}
+
+	if m.cacheEnabled && len(m.cacheOption.Group) > 0 {
+		keyGroup, errCache := cacheObj.Get(ctx, m.cacheOption.Group)
+		if errCache == nil {
+			keys := append(keyGroup.Strings(), cacheKey)
+			if errCache = cacheObj.Set(ctx, m.cacheOption.Group, keys, m.cacheOption.Duration); errCache != nil {
```
gqcn, line 180: The `Get` and `Set` function is not atomic, which will cause concurrent safety issue.

gqcn (changes requested):

database/gdb/gdb_model_cache.go
z5830865: You don't have to worry about concurrency security, and there's no problem with cache updates overwriting the original value

gqcn: @z5830865 Would it loose keys when there were multiple goroutines doing `Get` and `Set`?

gqcn (changes requested):

database/gdb/gdb_model_cache.go
gqcn: It is not good design using `Get/Set` to maintain the keys, as there's concurrent issue here.

gqcn (changes requested):

PR #2472, opened 2023-02-23
Protect RemoveValue from race condition

blinkbean:
When multiple gouroutines are trying to remove values from the same garray. After one routine calculate the index to remove, the other one could have removed the value before the calculated index, causing the index to shift and remove the wrong element. Making both the search and remove action atomic solve this problem.


gqcn (approved):

gqcn:
@blinkbean Thanks for your contribution! it's awesome!

PR #2488, opened 2023-03-03
add `Quick`  mode for gtimer

hinego:
This commit adds a new `Quick` bool to the timer code, which can be used to set the timer to quickly run without waiting for the first interval to be elapsed. This can be useful in certain scenarios where time-sensitive operations need to be performed immediately. 

此提交向 gtimer 添加了一个新的“Quick”布尔值，可用于将计时器设置为添加任务后立即执行，而无需等待第一个间隔结束。 这在某些需要立即执行时间敏感操作的场景中很有用。

PS：现在假设我添加一个30分钟执行的任务，则需要30分钟后才能第一次执行！

### DIFF
--- a/os/gtimer/gtimer_z_unit_entry_test.go
+++ b/os/gtimer/gtimer_z_unit_entry_test.go
@@ -60,6 +60,27 @@ func TestJob_Singleton(t *testing.T) {
 	})
 }
 
+func TestJob_SingletonQuick(t *testing.T) {
+	gtest.C(t, func(t *gtest.T) {
+		timer := gtimer.New(gtimer.TimerOptions{
+			Quick: true,
+		})
+		array := garray.New(true)
+		job := timer.Add(ctx, 5*time.Second, func(ctx context.Context) {
+			array.Append(1)
+			time.Sleep(10 * time.Second)
+		})
+		t.Assert(job.IsSingleton(), false)
+		job.SetSingleton(true)
+		t.Assert(job.IsSingleton(), true)
+		time.Sleep(250 * time.Millisecond)
+		t.Assert(array.Len(), 1)
+
+		time.Sleep(250 * time.Millisecond)
+		t.Assert(array.Len(), 1)
+	})
+}
+
 func TestJob_SetTimes(t *testing.T) {
 	gtest.C(t, func(t *gtest.T) {
 		timer := gtimer.New()
--- a/os/gtimer/gtimer_z_unit_timer_test.go
+++ b/os/gtimer/gtimer_z_unit_timer_test.go
@@ -103,6 +103,44 @@ func TestTimer_AddSingleton(t *testing.T) {
 	})
 }
 
+func TestTimer_AddSingletonWithQuick(t *testing.T) {
+	gtest.C(t, func(t *gtest.T) {
+		timer := gtimer.New(gtimer.TimerOptions{
+			Interval: 100 * time.Millisecond,
+			Quick:    true,
+		})
+		array := garray.New(true)
+		timer.AddSingleton(ctx, 5*time.Second, func(ctx context.Context) {
+			array.Append(1)
+			time.Sleep(10 * time.Second)
+		})
+		time.Sleep(250 * time.Millisecond)
+		t.Assert(array.Len(), 1)
+
+		time.Sleep(500 * time.Millisecond)
+		t.Assert(array.Len(), 1)
+	})
+}
+
+func TestTimer_AddSingletonWithoutQuick(t *testing.T) {
+	gtest.C(t, func(t *gtest.T) {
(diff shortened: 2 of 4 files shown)


gqcn:
@hinego I've thought about it. Is it better to call your timer function and then add it to timer?

hinego:
> @hinego I've thought about it. Is it better to call your timer function and then add it to timer?

I see what you mean, but its need extra coding like this
```
var timer = gtimer.New()

func test(ctx context.Context) {
	log.Println("test")
	if !added {
		timer.AddSingleton(xxx)
	}
	//or
	timer.AddOnce(xxx)
}
```
However, by my method,only need set the `gtimer.TimerOptions` dont need more code.

Like this
```
var timer = gtimer.New(gtimer.TimerOptions{
	Interval: 100,
	Quick:    true,
})

func init() {
	timer.AddSingleton(context.TODO(), 1*time.Second, test)
	timer.AddSingleton(context.TODO(), 1*time.Second, test2)
	timer.AddSingleton(context.TODO(), 1*time.Second, test3)
}
func test(ctx context.Context) {
	log.Println("test")
}
func test2(ctx context.Context) {
	log.Println("test")
}
func test3(ctx context.Context) {
	log.Println("test")
}
```

so...

gqcn:
@hinego Please add associated unit testing case for your updates.

hinego:
> @hinego Please add associated unit testing case for your updates.



> @hinego Please add associated unit testing case for your updates.

added

os/gtimer/gtimer_z_unit_entry_test.go
```diff
@@ -60,6 +60,27 @@ func TestJob_Singleton(t *testing.T) {
...
+		})
+		array := garray.New(true)
+		job := timer.Add(ctx, 5*time.Second, func(ctx context.Context) {
+			array.Append(1)
+			time.Sleep(10 * time.Second)
+		})
+		t.Assert(job.IsSingleton(), false)
+		job.SetSingleton(true)
+		t.Assert(job.IsSingleton(), true)
+		time.Sleep(250 * time.Millisecond)
```
gqcn, line 76: If `Quick`, is it better no using sleep after adding but directly do the asserting?

os/gtimer/gtimer_z_unit_timer_test.go
```diff
@@ -103,6 +103,44 @@ func TestTimer_AddSingleton(t *testing.T) {
...
+		timer := gtimer.New(gtimer.TimerOptions{
+			Interval: 100 * time.Millisecond,
+			Quick:    true,
+		})
+		array := garray.New(true)
+		timer.AddSingleton(ctx, 5*time.Second, func(ctx context.Context) {
+			array.Append(1)
+			time.Sleep(10 * time.Second)
+		})
+		time.Sleep(250 * time.Millisecond)
```
gqcn, line 117: As above.

gqcn (changes requested):

os/gtimer/gtimer_z_unit_entry_test.go
hinego:
```

// loop starts the ticker using a standalone goroutine.
func (t *Timer) loop() {
	go func() {
		var (
			currentTimerTicks   int64
			timerIntervalTicker = time.NewTicker(t.options.Interval)
		)
		defer timerIntervalTicker.Stop()
		for {
			select {
			case <-timerIntervalTicker.C:
				// Check the timer status.
				switch t.status.Val() {
				case StatusRunning:
					// Timer proceeding.
					if currentTimerTicks = t.ticks.Add(1); currentTimerTicks >= t.queue.NextPriority() {
						t.proceed(currentTimerTicks)
					}

				case StatusStopped:
					// Do nothing.

				case StatusClosed:
					// Timer exits.
					return
				}
			}
		}
	}()
}

```

Because it is not really executed immediately, but as soon as possible, because it may need to wait for the delay time of `timerIntervalTicker`

并不是添加后就立即执行

而是需要等待添加后被轮询到后才会被执行

假设：
```
var timer = gtimer.New(gtimer.TimerOptions{
	Interval: 10000, //10秒
	Quick:    true,
})
```

那么它可能最多延迟10秒才会被执行，而默认设定的是100 即100ms

即最多延迟100ms

hinego:
今天用到gtimer.once 时发现一个问题，如果quic时once这种就不适合了

所以要么，再适配一下，把once时不受quick的影响，要么once时不要使用quick模式

gqcn (approved):

gqcn:
> 今天用到gtimer.once 时发现一个问题，如果quic时once这种就不适合了
> 
> 所以要么，再适配一下，把once时不受quick的影响，要么once时不要使用quick模式

I think it needs no worry about once.

PR #2512, opened 2023-03-09
support microseconds resolution for created_at/updated_at/deleted_at

WesleyWu:
We often need microseconds resolution for timestamp/datetime columns, especially when we sort records by timestamp column and do some batch processing.

The current implementation in GF uses String() representation of gtime/time for insert/update operations, thus truncating the time precision to second.


WesleyWu:
finally get mssql driver test work.

WesleyWu:
MySQL 5.7 does not recoginze +07:00 in timestamp string, while MySQL 8.0 can.

contrib/drivers/dm/dm.go
```diff
@@ -11,19 +11,19 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
-	"net/url"
```
houseme, line 14: 这里不需要修改！

gqcn: @WesleyWu He means the imports sequence should not be changed: the std packages were put above, the others below.

gqcn (changes requested):

contrib/drivers/dm/dm.go
WesleyWu: sure

WesleyWu (commented):
restore the default organization of imports

gqcn (approved):

houseme:
> restore the default organization of imports

The unit test needs to add a test that is not six decimal places to verify that it can be written properly.

database/gdb/gdb_model_delete.go
```diff
@@ -63,7 +63,7 @@ func (m *Model) Delete(where ...interface{}) (result sql.Result, err error) {
 			Table:     m.tables,
 			Data:      fmt.Sprintf(`%s=?`, m.db.GetCore().QuoteString(fieldNameDelete)),
 			Condition: conditionStr,
-			Args:      append([]interface{}{gtime.Now().String()}, conditionArgs...),
+			Args:      append([]interface{}{gtime.Now().Local().Time.Format("2006-01-02 15:04:05.999999")}, conditionArgs...),
```
gqcn, line 66: Is it better using time object like `gtime.Now()` instead of time string？If so, you need no worry about time format.

database/gdb/gdb_model_update.go
```diff
@@ -73,7 +73,7 @@ func (m *Model) Update(dataAndWhere ...interface{}) (result sql.Result, err erro
 		// Automatically update the record updating time.
 		if fieldNameUpdate != "" {
 			if fieldNameUpdate != "" && !gstr.Contains(updates, fieldNameUpdate) {
-				updates += fmt.Sprintf(`,%s='%s'`, fieldNameUpdate, gtime.Now().String())
+				updates += fmt.Sprintf(`,%s='%s'`, fieldNameUpdate, gtime.Now().Local().Time.Format("2006-01-02 15:04:05.999999"))
```
gqcn, line 76: Is it better using time object like `gtime.Now()` instead of time string？like:
```go
updates += fmt.Sprintf(`,%s=?`, fieldNameUpdate)
conditionArgs = append(conditionArgs, gtime.Now())
```

gqcn (changes requested):

database/gdb/gdb_model_update.go
WesleyWu: Yeah. this solution eliminates the conversion from time to string.

WesleyWu:
现在应该比较完美了

houseme (approved):

houseme (approved):

WesleyWu:
> The unit test needs to add a test that is not six decimal places to verify that it can be written properly.

test added.

gqcn:
@WesleyWu Awesome!

gqcn (approved):

PR #2600, opened 2023-04-25
fix some comments

cuishuang:


net/gclient/gclient_z_example_test.go
```diff
@@ -529,7 +529,7 @@ func ExampleClient_GetVar() {
 // http proxy server listening on `127.0.0.1:1081`
 // socks5 proxy server listening on `127.0.0.1:1080`
 func ExampleClient_SetProxy() {
-	// connect to a http proxy server
+	// connect to an http proxy server
```
gqcn, line 532: Is it `a` correctly?

net/gclient/gclient_z_example_test.go
```diff
@@ -541,7 +541,7 @@ func ExampleClient_SetProxy() {
 	fmt.Println(err != nil)
 	resp.Close()
 
-	// connect to a http proxy server which needs auth
+	// connect to an http proxy server which needs auth
```
gqcn, line 544: As above.

net/ghttp/ghttp_server_router_group.go
```diff
@@ -200,52 +200,52 @@ func (g *RouterGroup) Map(m map[string]interface{}) {
...
 }
 
-// TRACE registers a http handler to give the route pattern and the http method: TRACE.
+// TRACE registers an http handler to give the route pattern and the http method: TRACE.
 func (g *RouterGroup) TRACE(pattern string, object interface{}, params ...interface{}) *RouterGroup {
 	return g.Clone().preBindToLocalArray(groupBindTypeHandler, "TRACE:"+pattern, object, params...)
 }
 
-// REST registers a http handler to give the route pattern according to REST rule.
+// REST registers an http handler to give the route pattern according to REST rule.
```
gqcn, line 248: As above.

gqcn (changes requested):

net/gclient/gclient_z_example_test.go
gqcn: > an http

OK, then.

gqcn (approved):

PR #2613, opened 2023-04-28
fix #2457 support for generic type in router function

### LINKED ISSUE #2457: need support for generic type in router function
### 1. What version of `Go` and system type/arch are you using?
go 1.19

### 2. What version of `GoFrame` are you using?
v2.3.2

### 4. What did you do?
```go
type ListMessageReq struct {
	g.Meta    `path:"/list" method:"get"`
	StartTime int64
	EndTime   int64
}
type ListMessageRes struct {
	g.Meta
	Title   string
	Content string
}
type BaseRes[T any] struct {
	g.Meta
	Code int
	Data T
	Msg  string
}
func (c *cMessage) List(ctx context.Context, req *v1.ListMessageReq) (res *v1.BaseRes[*v1.ListMessageRes], err error) {
	return res, err
}

```
报错日志：
[FATA] invalid struct naming for response: defined as "*v1.BaseRes[*xxx/api/v1.ListMessageRes]", but it should be named with "Res" suffix like "XxxRes"

**是不是要考虑泛型情况下，返回值名称校验？**

bobzh39:


net/ghttp/ghttp_server_service_handler.go
```diff
@@ -182,8 +182,13 @@ func (s *Server) checkAndCreateFuncInfo(f interface{}, pkgPath, structName, meth
 			return
 		}
 
+		req := reflectType.In(1).String()
```
gqcn, line 185: `reqTypeName`

net/ghttp/ghttp_server_service_handler.go
```diff
@@ -192,8 +197,13 @@ func (s *Server) checkAndCreateFuncInfo(f interface{}, pkgPath, structName, meth
 			return
 		}
 
+		res := reflectType.Out(0).String()
```
gqcn, line 200: `resTypeName`

gqcn (changes requested):

gqcn:
@bobzh39 Closed as inactivity for too long.

PR #2621, opened 2023-05-06
Correct the download address for CLI on Windows.

iciness:
Correct the download address for CLI on Windows.


hailaz:
谢谢，在以下pr修复 https://github.com/gogf/gf/pull/2622

PR #2633, opened 2023-05-08
fix: genpbentity type error, []byte => bytes

scoolor:
fix gf gen pbentity, []byte type error


gqcn:
@scoolor `go fmt` failed.

gqcn (approved):

PR #2658, opened 2023-05-19
feat(glog): add log rotation support for short-running process

xuyue86:
1.add file pattern matching detection for log rotation and backup file limitation:
before this commit:
log rotation is not working on short-running process

2.backup file limitation will working on all files matching pattern:

before this commit:
access-{y-m-d}.log will generate log like:
access-2023-05-19.log
access-2023-05-18.log

the limitation only working on rotation files like "access-2023-05-19.xxxxxxxxxxxx.log" , not counting the "access-2023-05-18.xxxxxxxxxxxxx.log"


gqcn:
@xuyue86 你这样改动确实能解决日志文件滚动的问题，但会额外引发性能问题，因为每一次写日志都会触发滚动逻辑。

xuyue86:
We have this condition，I think it will only triggered the first time when calling print.
`if !l.config.rotatedHandlerInitialized.Val() && l.config.rotatedHandlerInitialized.Cas(false, true) {
			l.rotateChecksTimely(ctx)
			intlog.Printf(ctx, "logger rotation initialized: every %s", l.config.RotateCheckInterval.String())
}`

os/glog/glog_logger_rotate.go
```diff
@@ -137,6 +138,11 @@ func (l *Logger) rotateChecksTimely(ctx context.Context) {
 		intlog.Errorf(ctx, `%+v`, err)
 	}
 	intlog.Printf(ctx, "logging rotation start checks: %+v", files)
+	// get file name regex pattern
+	// access-{y-m-d}-test.log => access-$-test.log => access-\$-test\.log => access-(.+?)-test\.log
+	fileNameRegexPattern, _ := gregex.ReplaceString(`{.+?}`, "$", l.config.File)
+	fileNameRegexPattern = gregex.Quote(fileNameRegexPattern)
+	fileNameRegexPattern = strings.ReplaceAll(fileNameRegexPattern, "\\$", "(.+?)")
```
gqcn, line 145: `strings.ReplaceAll` or `gregex.ReplaceString`?

gqcn (changes requested):

os/glog/glog_logger_rotate.go
xuyue86: `fileNameRegexPattern, _ := gregex.ReplaceString(`{.+?}`, "$", l.config.File)`   
using regex to replace "access-{y-m-d}-test.log" to "access-$-test.log".  The dollar symbol ($) is only a placeholder.
`fileNameRegexPattern = gregex.Quote(fileNameRegexPattern)` 
using regex quote to replace "access-$-test.log" to "access-\$-test\.log". After this replacement, the "fileNameRegexPattern" is only a normal string, not regex.
`fileNameRegexPattern = strings.ReplaceAll(fileNameRegexPattern, "\\$", "(.+?)")` 
using string replacement to change "access-\$-test\.log" to "access-(.+?)-test\.log", now, "fileNameRegexPattern" become real regex.

above commit changes are for one purpose: change gf's date format {y-m-d} to regex (.+?).

gqcn (approved):

PR #2695, opened 2023-06-09
feature:add CreateIntSliceFrom function

anyanfei:


gqcn:
@anyanfei Thanks for your contribution, but it is duplicated with `garray.NewIntArraySize`.

Issue #2677, opened 2023-06-01
neverDoneCtx forbids Done,  the signal of client disconnect to server can't be pass down to gf request context.


dukepan2005:
### 1. What version of `Go` and system type/arch are you using?
go version go1.20.3 darwin/amd64

### 2. What version of `GoFrame` are you using?
GoFrame Version: v2.4.1

### 3. Can this issue be re-produced with the latest release?
Yes

### 4. What did you do?

```
func(h *handler) Hander(cox context.Context, req *api.xxxxReq) (res *api.xxxxRes, err error) {
  go func(ctx context.Context){
    for {
      select {
      case <- ctx.Done():
        return
      default:
        g.Dump("default case")
      }
    }
  }(cox)

  .....
}
```

### 5. What did you expect to see?

When client close connect, we expect this goroutine should be exit.

### 6. What did you see instead?
this goroutine don't exist until server exit.

I checked source code, it's seems strange.

```
// neverDoneCtx never done.
type neverDoneCtx struct {
	context.Context
}
// Done forbids the context done from parent context.
func (*neverDoneCtx) Done() <-chan struct{} {
	return nil
}
```

Why do we forbids?

check document:
https://pkg.go.dev/net/http@go1.20.4#CloseNotifier

 If we forbids the context done from parent context. Maybe It has lost its original purpose in golang, and  not idiomatic.

WesleyWu:
I met the similar problem, but I can't be sure the problem is related to neverDownCtx.

I wrote a http handler to send stream (sse) to client. Sometimes the client may close because of timeout. The server needs to detect the client disconnection and stop send more events to it.

server code:
```
package main

import (
	"encoding/json"
	"time"

	"github.com/gogf/gf/v2/frame/g"
	"github.com/gogf/gf/v2/net/ghttp"
	"github.com/gogf/gf/v2/util/gconv"
)

type SampleMessage struct {
	Name  string
	Value string
}

func main() {
	s := g.Server()
	s.BindHandler("/stream", handleStream)
	s.SetPort(8080)
	s.Run()
}

func handleStream(r *ghttp.Request) {
	ctx := r.Context()

	r.Response.Header().Set("Content-Type", "text/event-stream")
	r.Response.Header().Set("Cache-Control", "no-cache")
	r.Response.Header().Set("Connection", "keep-alive")

	// 参考规范
	// https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events
	for i := 0; i < 100; i++ {
[...]

PR #2710, opened 2023-06-15
fix issue #2677 the signal of client disconnect to server ignored

WesleyWu:
When constructing a new ghttp.Request, the context should be set as the underlying context of wrapped http.Request.

### DIFF
--- a/net/ghttp/ghttp_request.go
+++ b/net/ghttp/ghttp_request.go
@@ -84,6 +84,7 @@ func newRequest(s *Server, r *http.Request, w http.ResponseWriter) *Request {
 		Request:       r,
 		Response:      newResponse(s, w),
 		EnterTime:     gtime.TimestampMilli(),
+		context:       r.Context(),
 		originUrlPath: r.URL.Path,
 	}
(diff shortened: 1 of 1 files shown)


WesleyWu:
The failed checks have nothing to do with the commits. How can I re-run the test?

dukepan2005:
Why should add new context field  to ghttp.Request?
Request has two methods in ghttp/ghttp_request_param_ctx.go, Context() and GetCtx()

If you init context in newRequest directly, the logic in Context() is useless.

So,  we should understand why forbids the context manually done in Context() method before submit PR.

WesleyWu:
> Why should add new context field to ghttp.Request? Request has two methods in ghttp/ghttp_request_param_ctx.go, Context() and GetCtx()
> 
> If you init context in newRequest directly, the logic in Context() is useless.
> 
> So, we should understand why forbids the context manually done in Context() method before submit PR.

I think neverDoneCtx context did not prevent the underlying context from triggering Done().
The reason of context not triggering Done() is because the real context was not passed as underlying context of the neverDoneCtx.

Issue #2719, opened 2023-06-25
`gjson.New(nil)` would be convert to empty string in db.Insert


niluan304:
### 1. ENV
```
go: go1.20 windows/amd64
gf: v2.4.4 (now latest release)
gf mysql driver: v2.4.4
mysql: 5.7.37
```

### 2. What did you do?

```go
gtest.C(t, func(t *gtest.T) {
    type Json struct {
        Id              int
        JsonEmptyString *gjson.Json
        JsonNil         *gjson.Json
        JsonNull        *gjson.Json
    }
    data := Json{
        Id:              1,
        JsonEmptyString: gjson.New(""),
        JsonNil:         gjson.New(nil),
        JsonNull:        gjson.New(struct{}{}),
    }
    _, err = db.Insert(ctx, tableName, data)
    t.AssertNil(err)
})
```

run and get error:

```log
1. INSERT INTO `nil1687704860540134700`(`id`,`json_empty_string`,`json_nil`,`json_null`) VALUES(1,'','','null') 
2. Error 3140: Invalid JSON text: "The document is empty." at position 0 in value for column 'nil1687704860540134700.json_empty_string'.
```

Reproducible code:

https://github.com/niluan304/gf/blob/27e1092f2217bffb79cd944c3d7259458666a28d/contrib/drivers/mysql/mysql_core_test.go#L261-L304

### 3. What did you expect to see?

`gjson.New('')` and ``gjson.New(nil)`` should  be convert to `'null'` but `''`

PR #2720, opened 2023-06-25
fix issue in converting value to nil when calling IsNil return ture for gdb.Core

niluan304:
```go
// iString is the type assert api for String.
type iString interface {
	String() string
}

// iNil if the type assert api for IsNil.
type iNil interface {
	IsNil() bool
}
```
check whether `value` implements the `iNil` interface, before `iString` interface.

fix issue in converting `gjson.New('')` and `gjson.New(nil)` to `nil`, related Issues: #2719


gqcn:
@niluan304 Awesome!

PR #2741, opened 2023-07-05
add endpoints configuration for `ghttp.Server`

gqcn:


net/ghttp/ghttp.go
```diff
@@ -101,7 +101,7 @@ type (
 	}
 
 	// ServerStatus is the server status enum type.
-	ServerStatus int
+	ServerStatus = int
```
houseme, line 104: ?

gqcn: That means type `ServerStatus` is alias of std type `int`, called `alias type`. The `alias type` cannot be a receiver, which means there cannot be function definitions on it.

houseme (approved):

PR #2762, opened 2023-07-13
fix g.Wait not support OS Signal

xyqweb:
fix g.Wait not support OS Signal


xyqweb (commented):
fix g.Wait not support OS Signal

PR #2784, opened 2023-07-19
add `NeverDone` function for package `gctx` that creates and returns a never done ctx

gqcn:


contrib/drivers/oracle/oracle.go
```diff
@@ -19,9 +18,10 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/gogf/gf/v2/util/gutil"
 	gora "github.com/sijms/go-ora/v2"
 
+	"github.com/gogf/gf/v2/util/gutil"
+
```
houseme, line 24: 这行多余

gqcn: @houseme done

houseme (approved):

Issue #2782, opened 2023-07-19
[bug] ORM builder with OmitEmpty generate wrong SQL


Agzer0:
### 1. What version of `Go` and system type/arch are you using?

go1.20.3

### 2. What version of `GoFrame` are you using?

v2.4.0

### 3. Can this issue be re-produced with the latest release?
Yes

### 4. What did you do?

table schema:
```sql
CREATE TABLE `doctor` (
  `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
  `mobile` varchar(15) NOT NULL,
  `name` varchar(50) NOT NULL ,
  `status` tinyint(1) unsigned NOT NULL DEFAULT '1' ,
  `created_at` datetime NOT NULL,
  `updated_at` datetime NOT NULL,
  `deleted_at` datetime DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
```

```go
// dao.Doctor is gf generate dao struct
m := dao.Doctor.Ctx(gctx.New())

var doctor entity.Doctor
_ = m.Where("mobile", "").
		Where(m.Builder().Where("id", 1).WhereOrLike("name", "foo")).
		Where("status", 2).
		OmitEmpty().Scan(&doctor)
// expect sql: 
// SELECT `id` FROM `doctor` WHERE ((`id`=1) OR (`name` LIKE 'foo')) AND (`status`=2) AND `deleted_at` IS NULL LIMIT 1

// But the generated SQL below is wrong:
// The first statement is filtered out, second is a nested statement, but it's been expanded to the level of the third statement
// SELECT `id` FROM `doctor` WHERE ((`id`=1) OR (`name` LIKE 'foo') AND (`status`=2)) AND `deleted_at` IS NULL LIMIT 1
```

PR #2787, opened 2023-07-20
fix: orm generate sql wrong, issues #2782

Agzer0:
fix issues #2782 : When ORM is constructing multiple SQL conditions, the OmitEmpty option is used. If the first condition is empty conditions and the second condition is the compound condition. At this time, when there is still a third condition, the generated SQL is wrong.


gqcn:
@Agzer0 Thanks for your contribution, but please commit associated unit testing case for this update. Please add the unit testing case under driver: https://github.com/gogf/gf/tree/master/contrib/drivers/mysql

Agzer0:
okay, i`ve added testing case

gqcn (approved):

PR #2893, opened 2023-08-23
fix: add a expiration clear callback function to gcache

ok-dok:
The gcache needs set a callback function when the key and value in cache is expired to clear.

In some business scenarios, it use gcache in memory to cache some stateful values,  such as a client that keep a connection, 
the callback function is useful to close it and execute some clear operations when cached value is expired.

I add a method "SetOnExpiredClearCallBack(ctx context.Context, callback func(key interface{}, value *gvar.Var))" in Adapter interface and the implement AdapterMemory struct.

Can you merge it?

### DIFF
--- a/os/gcache/gcache_adapter.go
+++ b/os/gcache/gcache_adapter.go
@@ -139,4 +139,8 @@ type Adapter interface {
 
 	// Close closes the cache if necessary.
 	Close(ctx context.Context) error
+
+	// SetOnExpiredClearCallBack set a callback function that used when clear the expired key and value.
+	// This callback function will execute after clear the expired key and value
+	SetOnExpiredClearCallBack(ctx context.Context, callback func(key interface{}, value *gvar.Var))
 }
--- a/os/gcache/gcache_adapter_memory.go
+++ b/os/gcache/gcache_adapter_memory.go
@@ -8,7 +8,9 @@ package gcache
 
 import (
 	"context"
+	"fmt"
 	"math"
+	"os"
 	"time"
 
 	"github.com/gogf/gf/v2/container/glist"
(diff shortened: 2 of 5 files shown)


gqcn:
@ok-dok 
Hi, thanks for your contribution to GoFrame! 

But it might not be a good idea that sets a global expire callback function for all values in the cache, as there might be different types of values in the cache not only the connection objects. And also, the expiration callback feature would slow down/block the expiration logic of the cache and cause unexpected issue.

gqcn:
@ok-dok Is it better add an interface that would be called when the object, which implements the interface, in cache is expired? For example:
```go
type IExpire interface{
    Expire()
}
```

ok-dok:
> @ok-dok Hi, thanks for your contribution to GoFrame!
> 
> But it might not be a good idea that sets a global expire callback function for all values in the cache, as there might be different types of values in the cache not only the connection objects. And also, the expiration callback feature would slow down/block the expiration logic of the cache and cause unexpected issue.
> 
> It is not a good proposal for the cache manager maintaining the stateful objects, as I see.

yes, I have noted your reply last mounth. But I think set a callback function to use when excuting clear action is useful. The SetOnExpiredClearCallBack function I added used a go routine and executed with panic catch recover after cleared, so it won't slow down the execution speed of cache cleaning.

`type IExpire interface{
    Expire()
}`

Adding an interface like the example looks like a good idea, but I'm am worried that it may involve the transformation of third-party class libraries. When using third-party struct, It may not implements this interface, for continue using gcache, I need to write a new struct that embedding the source struct and implements the IExpire interface.

ok-dok:
I found a issue https://github.com/gogf/gf/issues/527 that discuss this question, but it's been almost three years now and there's still no discussion on the outcome, I think it's necessary to add a cache cleared event listener in some cases.

Issue #2665, opened 2023-05-25
*time. Time is inconsistent with time.Time conversion


Rehtt:
### 1. What version of `Go` and system type/arch are you using?

go version go1.18.4 windows/amd64

### 2. What version of `GoFrame` are you using?

VERSION = "v2.4.1"

### 4. What did you do?

```
	var a struct {
		T time.Time `json:"t"`
	}
	var tmp = map[string]any{"t": "2023-04-15 19:10:00 +0800 CST"}
	err := gconv.Struct(tmp, &a)
	if err != nil {
		panic(err)
	}
	fmt.Println(a)

#  {2023-04-15 19:10:00 +0800 CST}
```

```
	var a struct {
		T *time.Time `json:"t"`
	}
	var tmp = map[string]any{"t": "2023-04-15 19:10:00 +0800 CST"}
	err := gconv.Struct(tmp, &a)
	if err != nil {
		panic(err)
	}
	fmt.Println(a)

# panic: error binding value to attribute "T": parsing time "2023-04-15 19:10:00 +0800 CST" as "2006-01-02T15:04:05Z07:00": cannot parse " 19:10:00 +0800 CST" as "T"
```

```
	var a struct {
		T *time.Time `json:"t"`
	}
	var tmp = map[string]any{"t": "2006-01-02T15:04:05Z07:00"}
	err := gconv.Struct(tmp, &a)
	if err != nil {
		panic(err)
	}
	fmt.Println(a)

# panic: error binding value to attribute "T": parsing time "2006-01-02T15:04:05Z07:00": extra text: "07:00"
```

PR #2915, opened 2023-08-30
fix: multiple interfaces cause the original type to be inaccessible

wlynxg:
fix: https://github.com/gogf/gf/issues/2665

Since v is of `interface{}` type, when v is assigned a value, the type of v is `interface{}|time.Time`, the function return type is interface{}, and the value type at this time is `interface{}|*interface{}|time .Time`, this will result in the original type `time.Time` not being available outside.

### DIFF
--- a/util/gconv/gconv_convert.go
+++ b/util/gconv/gconv_convert.go
@@ -194,7 +194,7 @@ func doConvert(in doConvertInput) (convertedValue interface{}) {
 		}
 		return Time(in.FromValue)
 	case "*time.Time":
-		var v interface{}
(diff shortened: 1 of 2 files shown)


gqcn:
@wlynxg It's awesome! Would you please add associated unit testing case to cover this issue https://github.com/gogf/gf/issues/2665 ?

wlynxg:
> @wlynxg It's awesome! 💓 Would you please add associated unit testing case to cover this issue #2665 ?

👌

gqcn (approved):

PR #2962, opened 2023-09-14
It is up to the developer to choose whether to enable chain-safe mode for the generated Dao

wsy998:
I have added a configuration to allow developers to choose whether to use Safe when generating DAOs. However, since most developers may not need to worry about this, I have set the default value to true. When it's set to false, Safe mode is not used. This may be more friendly for developers who are sensitive to speed or want more control over cloning.


gqcn:
@wsy998 You cannot use unsafe dao object as it has concurrent safety issue.

wsy998:
> @wsy998 Thanks for you contribution, but you cannot use unsafe dao object as it has concurrent safety issue when dao object is used in multiple goroutines.

那为什么当时要将数据库设计成非链路安全呢，而且我已经默认将Safe模式设为true了

wsy998:
> @wsy998 Thanks for you contribution, but you cannot use unsafe dao object as it has concurrent safety issue when dao object is used in multiple goroutines.

或是我提供两个方法，一个提供Safe模式一个不是Safe模式

PR #2969, opened 2023-09-15
feat: Preserve original file permissions & Default copy permissions c…

wlynxg:
1. DefaultPermCopy permissions changed to 755, 777 permission is too great;
2. CopyOption adds an option to copy source file permissions;
3. Added unit tests related to CopyOption.

### DIFF
--- a/os/gfile/gfile_copy.go
+++ b/os/gfile/gfile_copy.go
@@ -17,9 +17,16 @@ import (
 
 // CopyOption is the option for Copy* functions.
 type CopyOption struct {
-	Sync         bool        // Auto call file sync after source file content copied to target file.
-	PreserveMode bool        // Preserve the mode of the original file to the target file.
-	Mode         os.FileMode // Destination created file mode. The default file mode is DefaultPermCopy.
+	// Auto call file sync after source file content copied to target file.
+	Sync bool
+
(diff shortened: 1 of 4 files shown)


os/gfile/gfile_copy.go
```diff
@@ -17,8 +17,9 @@ import (
 
 // CopyOption is the option for Copy* functions.
 type CopyOption struct {
-	Sync bool        // Auto call file sync after source file content copied to target file.
-	Mode os.FileMode // Destination created file mode. The default file mode is DefaultPermCopy.
+	Sync         bool        // Auto call file sync after source file content copied to target file.
+	PreserveMode bool        // Preserve the mode of the original file to the target file.
+	Mode         os.FileMode // Destination created file mode. The default file mode is DefaultPermCopy.
```
gqcn, line 22: It should be described the relation between `PreserveMode` and `Mode`, like:
```go
PreserveMode bool        // Preserve the mode of the original file to the target file. If true, the Mode attribute will make no sense.
Mode         os.FileMode // Destination created file mode. The default file mode is DefaultPermCopy if PreserveMode is false.
```

gqcn (changes requested):

gqcn (approved):

PR #2981, opened 2023-09-22
Let grpc server can fix ip and random port.

joy999:
Let grpc server can set one fixed ip and random port.


gqcn:
@joy999 Please use `ip:0` to use random port with given listening ip.

PR #2985, opened 2023-09-26
Provide some common functions for gJson.

wsy998:
Such as gjson#GetString()


gqcn:
@wsy998 The committed functions are unnecessary. You can use `Get(xxx).Xxx` to retrieve value and convert it to certain type of value.

PR #2991, opened 2023-09-27
add direct service address support for grpc client

gqcn:


contrib/rpc/grpcx/grpcx.go
```diff
@@ -29,6 +29,7 @@ const (
 const (
 	defaultServerName        = `default`
 	configNodeNameGrpcServer = `grpc`
+	delimiterOfAddress       = ":"
```
houseme, line 32: `gsvc.EndpointHostPortDelimiter` 一样的功能

gqcn: done

houseme (approved):

PR #3048, opened 2023-10-13
fix:incorrect parameter passing causes captured errors to be lost

lazyironf4ur:


gqcn:
@lazyironf4ur Thanks for your contribution! ❤️

Issue #3099, opened 2023-10-26
Need fix: Can't convert gtime.Time to timestamppb now.


joy999:
### 1. What version of `Go` and system type/arch are you using?
go 1.20

### 2. What version of `GoFrame` are you using?
2.5.6

### 3. Can this issue be re-produced with the latest release?
yes

### 4. What did you do?
The converter can't work well now. It worked well before.
func(in gtime.Time) (*timestamppb.Timestamp, error) {
	return utility.GTimeToPbTime(&in), nil
}

### 5. What did you expect to see?
Let it work well.

### 6. What did you see instead?
It don't work.

PR #3102, opened 2023-10-26
fix issue #3099

joy999:


util/gconv/gconv_z_unit_converter_test.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/gogf/gf/v2/os/gtime"
 	"github.com/gogf/gf/v2/test/gtest"
 	"github.com/gogf/gf/v2/util/gconv"
+	"google.golang.org/protobuf/types/known/timestamppb"
```
gqcn, line 17: It will add extra dependent package to `go.mod`. Is it better move this unit testing case to example folder?

util/gconv/gconv_struct.go
```diff
@@ -377,6 +377,12 @@ func bindVarToStructAttr(structReflectValue reflect.Value, attrName string, valu
 	if empty.IsNil(value) {
 		structFieldValue.Set(reflect.Zero(structFieldValue.Type()))
 	} else {
+		// Try to call custom converter.
+		// Issue: https://github.com/gogf/gf/issues/3099
+		if ok, err := callCustomConverter(reflect.ValueOf(value), structFieldValue); ok {
```
gqcn, line 382: It'd be better to check that the `value` is already type of `reflect.Value` or not.

gqcn (changes requested):

util/gconv/gconv_z_unit_converter_test.go
joy999: I think maybe I can create a new struct instead of it.

util/gconv/gconv_struct.go
joy999: improved

PR #3109, opened 2023-10-28
Add the Protocol attribute to the Redis configuration

tianxing113:
The go-redis v9 has added support for RESP3 and communicates with the redis server using RESP3 by default. However, many cloud server manufacturers’ redis cloud services do not yet support RESP3. Therefore, when using the v9 client to connect to a redis server that does not support RESP3, some redis commands may fail to process the return value from the server, such as ‘zset range’. In order to be compatible with the RESP2 protocol, it is necessary to add configuration for the RESP version.

### DIFF
--- a/contrib/nosql/redis/redis.go
+++ b/contrib/nosql/redis/redis.go
@@ -61,6 +61,7 @@ func New(config *gredis.Config) *Redis {
 		WriteTimeout:    config.WriteTimeout,
 		MasterName:      config.MasterName,
 		TLSConfig:       config.TLSConfig,
+		Protocol:        config.Protocol,
 	}
 
 	var client redis.UniversalClient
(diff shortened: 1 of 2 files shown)


houseme (approved):

roufutuan:
这个PR大概什么时候能进版本？我们项目也遇到了这个问题，腾讯云的全线redis产品都只支持RESP2

hailaz (approved):

gqcn:
@tianxing113 Awesome! Thanks for your contribution!

gqcn:
> 这个PR大概什么时候能进版本？我们项目也遇到了这个问题，腾讯云的全线redis产品都只支持RESP2

It now merged into branch `master`, and will be release in next version. You can use `go get -u github.com/gogf/gf/v2@master` if you're urgent.

PR #3258, opened 2024-01-11
fix gf gen service

loveyfore:
Fix the issue of significant differences in the generated code every time when executing the 'gf gen service' command.

### DIFF
--- a/net/ghttp/ghttp_server_pprof.go
+++ b/net/ghttp/ghttp_server_pprof.go
@@ -26,7 +26,7 @@ const (
 // StartPProfServer starts and runs a new server for pprof.
 func StartPProfServer(port int, pattern ...string) {
 	s := GetServer(defaultPProfServerName)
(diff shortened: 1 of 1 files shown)


cmd/gf/internal/cmd/genservice/genservice.go
```diff
@@ -174,7 +175,7 @@ func (c CGenService) Service(ctx context.Context, in CGenServiceInput) (out *CGe
 		// Parse single logic package folder.
 		var (
 			// StructName => FunctionDefinitions
-			srcPkgInterfaceMap   = make(map[string]*garray.StrArray)
+			srcPkgInterfaceMap   = orderedmap.NewOrderedMap[string, *garray.StrArray]()
```
gqcn, line 178: Please use `gmap.ListMap` as the sorted map container.

gqcn (changes requested):

PR #3273, opened 2024-01-17
add example for ctx keys feature of package `glog`

gqcn:


example/os/log/ctx_keys/main.go
```diff
@@ -0,0 +1,14 @@
+package main
```
houseme, line 1: copyright

houseme (approved):

Issue #3325, opened 2024-02-25
gproc.signal enhance


Agzer0:
**Package that You wish to enhance**
signal funcs for`gproc`

**Enhancement description**
1. The Listen() method should be executed only once. In business encapsulation, signal listening might be called in different packages, and there could be multiple calls to Listen() (typically done in a goroutine), resulting in the handlers being registered and executed multiple times.

2. After the execution of Listen(), calling AddSigHandler() to add new custom signal handlers may fail to properly listen to the signals.

3. If Listen() is called in one goroutine, and then AddSigHandler()/AddSigHandlerShutdown() is called in another goroutine, it may lead to data race issues, primarily due to the signalHandlerMap, an internal variable of type map.

**Additional**

Agzer0:
I see that the 3rd point has been addressed in a later version

PR #3326, opened 2024-02-25
enhance: gproc.signal enhance #3325

Agzer0:


gqcn:
@Agzer0 Would you please give a brief description about your changes?

Agzer0:
Implemented the following changes for gogf#3325:

1. Used sync.Once to restrict Listen() so that it is executed only once, avoiding multiple executions of Listen() that could lead to repetitive triggering of handlers.
2. Used waitChan to ensure that Listen() remains blocking.
3. In AddSigHandlerShutdown() and AddSigHandler(), invoked signal.Notify() everytime so that even if signals are added after Listen(), they can still be registered and listened to.

os/gproc/gproc_z_signal_test.go
```diff
@@ -0,0 +1,97 @@
+package gproc
```
gqcn, line 1: Please add copyright comments in the top of source file.

gqcn (changes requested):

gqcn (approved):

PR #3420, opened 2024-03-22
enhance: use in-memory cache for soft field name and type

### LINKED ISSUE #3419: Cache Conflicts in `getSoftFieldNameAndType`
**What version of `Go` and system type/arch are you using?**

`go version go1.21.8 linux/amd64`

**What version of `GoFrame` are you using?**

`v2.6.4`

**Can this bug be re-produced with the latest release?**
yes

**What did you do?**

Create a database model with fields `created_at`, `updated_at`, `deleted_at`. Make all server instances to use the same redis cache. When these servers run a period of time (including many server restarts), goframe may update `deleted_at` automatically when updating records in database. Sometimes the generated SQL also contains `WHERE update_at IS NULL`.

**What did you expect to see?**
The field `deleted_at` should not be updated automatically in update operations.

**What did you see instead?**
It is updated.

**Possible Reasons :**
The cache in `getSoftFieldNameAndType` is never cleaned or expired.
https://github.com/gogf/gf/blob/313d9d138f96b0ed460d47684298a7fb26d3fd75/database/gdb/gdb_model_soft_time.go#L187
The cache key uses schema, table name and the address of slice `checkFiledNames`. If we are luck enough, the three slices here may use the same address on different machines at the same time in distributed scenarios. Because the cache is not cleaned, the cache can also be hit by mistake after server restarts. https://github.com/gogf/gf/blob/313d9d138f96b0ed460d47684298a7fb26d3fd75/database/gdb/gdb_model_soft_time.go#L76

**Possible Solutions:**
Use the content of `checkFiledNames` or hash values of the content as part of the cache key.

helloqiu:
fix #3419


database/gdb/gdb_model_soft_time.go
```diff
@@ -214,7 +219,7 @@ func (m *softTimeMaintainer) getSoftFieldNameAndType(
 			return
 		}
 	)
-	result, err := m.db.GetCache().GetOrSetFunc(ctx, cacheKey, cacheFunc, cacheDuration)
+	result, err := softFieldNameAndTypeCache.GetOrSetFunc(ctx, cacheKey, cacheFunc, cacheDuration)
```
gqcn, line 222: Is t better using `gcache` here instead of new cache object?

gqcn (changes requested):

database/gdb/gdb_model_soft_time.go
helloqiu: Thanks for the review. Sure, `gcache` use the default in-memory cache and it fits here.

gqcn (approved):

PR #3450, opened 2024-04-02
fix: `overwrite` typo in return error of package `gvalid`

phuonganhniie:


util/gvalid/gvalid_register.go
```diff
@@ -52,7 +52,7 @@ func RegisterRule(rule string, f RuleFunc) {
 	if customRuleFuncMap[rule] != nil {
 		intlog.PrintFunc(context.TODO(), func() string {
 			return fmt.Sprintf(
-				`rule "%s" is overwrotten by function "%s"`,
+				`rule "%s" is overwrite by function "%s"`,
```
gqcn, line 55: It might be `overwritten`...

gqcn (changes requested):

PR #3451, opened 2024-04-03
feat: add `Scan` method for incremental key retrieval in `gredis`

phuonganhniie:
## Overview
This pull request introduces the `Scan` method to the `GroupGeneric` struct in `gredis` package, providing an efficient and non-blocking approach to iterate over keys that match a given pattern in Redis. This method is designed to handle large datasets more gracefully than the `Keys` command, offering incremental retrieval without the risk of blocking the Redis server.

## Changes
- **`Scan` Method**: Implements the Redis `SCAN` command within the `GroupGeneric` struct. It accepts a cursor, pattern, and count, returning a slice of keys matching the pattern and the next cursor position for continued scanning.

## Why It's Needed
The `SCAN` command is essential for applications dealing with large data sets in Redis, where using the `Keys` command might lead to performance bottlenecks or blocking the server. By providing a `Scan` method, it easier and safer to get keys little by little, improving overall application responsiveness.

Available since Redis version: `2.8.0`

## Usage Example
Below is a simple example demonstrating how to use the new `Scan` method:

```go
var cursor uint64 = 0
var allKeys []string
var count int = 100

for {
    keys, nextCursor, err := redis.GroupGeneric().Scan(ctx, cursor, "*pattern*", count)
    if err != nil {
        log.Fatalf("Scan failed: %v", err)
    }
    allKeys = append(allKeys, keys...)
    if nextCursor == 0 {
        break
    }
    cursor = nextCursor
}

fmt.Println("Retrieved keys:", allKeys)
```

## Testing
Includes comprehensive unit tests covering:
- Successful key retrieval with various patterns.
- Proper handling of negative cursor and count values.
- Edge cases, such as empty patterns, pattern with no matches.

I'd love to hear your thoughts and feedback on this! Looking forward to your review.


phuonganhniie:
@gqcn Hi John, I'm looking forward to your review. If you need any information, pls don't hesitate contact me.

contrib/nosql/redis/redis_group_generic.go
```diff
@@ -170,6 +170,26 @@ func (r GroupGeneric) Keys(ctx context.Context, pattern string) ([]string, error
...
 
+// Scan executes a single iteration of the SCAN command, returning a subset of keys matching the pattern along with the next cursor position.
+// This method provides more efficient and safer way to iterate over large datasets compared to KEYS command.
+//
+// Users are responsible for controlling the iteration by managing the cursor.
+//
+// The `count` parameter advises Redis on the number of keys to return. While it's not a strict limit, it guides the operation's granularity.
+//
+// https://redis.io/commands/scan/
+func (r GroupGeneric) Scan(ctx context.Context, cursor uint64, pattern string, count int) ([]string, uint64, error) {
```
gqcn, line 181: As the command decribed in https://redis.io/commands/scan/ , the `MATCH/COUNT/TYPE` is an option, which would be better defined as an optional paremeter in this fuction, too.
And also, the first return value would be better defined as the cursor number.
图片

gqcn (changes requested):

hailaz (approved):

contrib/nosql/redis/redis_group_generic.go
phuonganhniie: Thank you for your insightful feedback. I've taken your suggestions and made the following adjustments:

- `Optional Parameters`: I've updated the implementation to reflect `MATCH/COUNT/TYPE` as truly optional parameters within the Scan function. This is achieved through the use of the `ScanOption` struct, where each field is a pointer, allowing callers to specify only the options they need without providing values for all parameters.

- Cursor as the first return value, and array of elements as the second return value. 

I appreciate your guidance and hope these changes address your concerns effectively. Looking forward to any further suggestions or approval.

database/gredis/gredis_redis_group_generic.go
```diff
@@ -60,3 +61,9 @@ type ExpireOption struct {
 	GT bool // GT -- Set expiry only when the new expiry is greater than current one
 	LT bool // LT -- Set expiry only when the new expiry is less than current one
 }
+
+type ScanOption struct {
+	Match *string // Match -- Specifies a glob-style pattern for filtering keys.
+	Count *int    // Count -- Suggests the number of keys to return per scan.
+	Type  *string // Type -- Filters keys by their data type. Valid types are "string", "list", "set", "zset", "hash", and "stream".
+}
```
gqcn, line 69: - Missing comments for public struct. 
_You might say there're no comments for some of the public interface methods and types either, but you see this is heavy work that we have not finished yet._
- Is it necessary using pointer as the option attribute? It's not convenient for developers to use it, as the developers would need convert the variable value to pointer before using this option.

gqcn (changes requested):

database/gredis/gredis_redis_group_generic.go
phuonganhniie: - I updated the comments for the `ScanOption` public struct. 
- About using pointer as the option attributes: Because if we don't use pointer, the `convertOptionToArgs` function will appends attributes with their zero value into `args` variable. The `Do` method will executes `Scan` command with all of values in the `args` and returns unexpected result. 

For example:
- Data in Redis:
```json
"firstname": "Jack",
"lastname":  "Stuntman",
"age":       35,
"nickname":  "Jumper"
```

- Run Test:
```Go
type ScanOption struct {
  Match string 
  Count int
  Type  string
}

func Test_GroupGeneric_Scan(t *testing.T) {
    gtest.C(t, func(t *gtest.T) {
        performScan := func(cursor uint64, option *ScanOption) ([]string, error) {
	        var allKeys = []string{}
	        for {
		        var nextCursor uint64
		        var keys []string
		        var err error
        
		        if option != nil {
			        nextCursor, keys, err = redis.Scan(ctx, cursor, *option)
		        } else {
			        nextCursor, keys, err = redis.Scan(ctx, cursor)
		        }
		        if err != nil {
			        return nil, err
		        }
        
		        allKeys = append(allKeys, keys...)
		        if nextCursor == 0 {
			        break
		        }
		        cursor = nextCursor
	        }
	        return allKeys, nil
        }
        
        optWithName := gredis.ScanOption{Match: "*name*", Count: 10}
        keysWithName, err := performScan(0, &optWithName)
        t.AssertNil(err)
        if len(keysWithName) < 3 {
	        t.Errorf("Expected at least 3 keys matching '*name*', found %d", len(keysWithName))
        }
    }
}
```

- Test result: 
```
Expected at least 3 keys matching '*name*', found 0
```

That means the `Scan` method also accepts the field `Type` with empty string instead of ignores it. So if we want to the `MATCH/COUNT/TYPE` is optional, I have to define those fields's data type is pointer. 

If you have the other solution, I really looking forward to learn from it.

gqcn: @phuonganhniie You can macnually convert the option to the args that will be passed to redis, for example:
```go
var args = make([]any, 0)
if usedOption.Match != "" {
    args = append(args, "MATCH", usedOption, Match)
}
if usedOption.Count > 0 {
    args = append(args, "COUNT", usedOption, Count)
}
if usedOption.Type != "" {
    args = append(args, "TYPE", usedOption, Type)
}
v, err := r.Operation.Do(ctx, "Scan", args...)
```

Or else, you can manually convert the public `ScanOption` to internal `doScanOption` using pointer attributes, which can be directly used in function `mustMergeOptionToArgs`, for example:
```go
type ScanOption struct {
	Match string // Match -- Specifies a glob-style pattern for filtering keys.
	Count int    // Count -- Suggests the number of keys to return per scan.
	Type  string // Type -- Filters keys by their data type. Valid types are "string", "list", "set", "zset", "hash", and "stream".
}

type doScanOption struct {
	Match *string 
	Count *int   
	Type  *string 
}
```

phuonganhniie: @gqcn Hi John, I've updated by transform the public `ScanOption` to internal `doScanOption` using pointer attributes. Please help me review it and thank you for your patience.

gqcn (approved):

PR #3472, opened 2024-04-10
fix: glog - empty content should has placeholder

primexiao:
When I pass values... to glog, if you totally ignore the empty values, I will be confused where is my params or if I made a  mistake ? Also, multi values may lead to out-of-order if no placeholder for empty values.


gqcn:
@primexiao Let's take `fmt.Printf("hello %s", "")` for example, why should the empty string be hiddenly converted to a space?

primexiao:
> @primexiao Let's take `fmt.Printf("hello %s", "")` for example, why should the empty string be hiddenly converted to a space?

But glog use " " as separator with each values by default, which I think it looks more like the behavior of  `strings.Join(values, ' ')`

PR #3483, opened 2024-04-12
fix: missing file closding when printing downloading percent of gf cli file for command `gf up`

testwill:


hailaz (approved):

cmd/gf/internal/utility/utils/utils_http_download.go
```diff
@@ -79,6 +79,8 @@ func doPrintDownloadPercent(doneCh chan int64, localSaveFilePath string, total i
 			if err != nil {
 				mlog.Fatal(err)
 			}
+			defer file.Close()
```
gqcn, line 82: It'd better move this file `Open` logic out side of for loop.

gqcn (changes requested):

gqcn (approved):

PR #3562, opened 2024-05-06
net/ghttp: conditional printing access log to stdout

chamhaw:
Added AccessLogStdoutDisabled toggle to disable access content logging into stdout and supported conditional access log to exclude endpoints such as /health, /metrics etc.


gqcn:
@chamhaw Hi, the logging content to stdout can be disabled though configuration `LogStdout`. The logging content should not be filtered by configuration, that is, the logging content should be complete or nothing but not be filtered manually.

PR #3583, opened 2024-05-16
driver/pgsql: optimize regex for version matching

yincongcyincong:
**Please ensure you adhere to every item in this list.**
every time function **Index** and **version** execute, these regexp are initialized. which situation reduce the performance.

### DIFF
--- a/cmd/gf/internal/cmd/cmd_build.go
+++ b/cmd/gf/internal/cmd/cmd_build.go
@@ -35,6 +35,8 @@ var (
 		nodeNameInConfigFile: "gfcli.build",
 		packedGoFileName:     "internal/packed/build_pack_data.go",
 	}
+
+	spaceRegex  = regexp.MustCompile(`\s+`)
 )
 
 type cBuild struct {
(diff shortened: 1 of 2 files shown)


cmd/gf/internal/cmd/cmd_build.go
```diff
@@ -190,7 +192,6 @@ func (c cBuild) Index(ctx context.Context, in cBuildInput) (out *cBuildOutput, e
 	}
 	// System and arch checks.
 	var (
-		spaceRegex  = regexp.MustCompile(`\s+`)
```
gqcn, line 193: It is OK here in CLI command, as it is executed once.

contrib/drivers/pgsql/pgsql_tables.go
```diff
@@ -32,6 +32,8 @@ WHERE
 ORDER BY
 	c.relname
 `
+
+	postgreRegex = regexp.MustCompile(`PostgreSQL (\d+\.\d+)`)
```
gqcn, line 36: It's better named as `versionRegex`, as it is used for version match.

gqcn (changes requested):

cmd/gf/internal/cmd/cmd_build.go
yincongcyincong: but this regex use in a **for**. i this it will be executed  several times.

contrib/drivers/pgsql/pgsql_tables.go
yincongcyincong: i'll change the name

gqcn (approved):

PR #3586, opened 2024-05-17
cmd/gf: Add a safe option to the Ctx func in TemplateGenDaoIndexContent

CyJaySong:
cmd/gf: Add a safe option to the Ctx func in TemplateGenDaoIndexContent


gqcn:
@CyJaySong It looks that there's no necessary adding an optional `safe` parameter.

CyJaySong:
@gqcn When calling Ctx functions, Ide's intelligent prompts can be used to remind developers to determine whether they need to use`safe`

PR #3618, opened 2024-05-30
cmd/gf: use automatic generating package names for go files of `dao/do/entity` instead of constant package names

Insua:
When employing the CLI to instantiate the generation of dao, one customarily specifies parameters such as daoPath, doPath, and entityPath. These parameters determine the file paths for the generated Go files, while the package names default to dao, do, and entity. To address this, an additional configuration parameter, withPathPackage, has been introduced. If withPathPackage is set to true, the package names will be defined according to the respective paths specified by daoPath, doPath, and entityPath.


Insua:
这个我也关了吧，有多个commit，我重新提交一下pr

Insua:
**I closed this PR. It has multiple commits. Just use a new PR.**
----------------------------------------
这个pr我关闭了，这个也有多次commit，用新的pr吧
#3639

PR #3637, opened 2024-06-07
drivers/pgsql:pgsql returning id should use quotation marks,when primary key is capital

Insua:
When the field name in pgsql is uppercase, the field must be enclosed in quotation marks; otherwise, an error will occur indicating that the field cannot be found. Therefore, a modification was made to change it to RETURNING "ID" in this style.
------------------------------
当pgsql的字段名大写时，必须把字段用"给引用起来，否则会报错，提示找不到字段，所以作出修改，改成了 RETURNING "ID"，这种样式。


cmd/gf/internal/cmd/gendao/gendao.go
```diff
@@ -86,6 +86,7 @@ CONFIGURATION SUPPORT
 	CGenDaoBriefClear             = `delete all generated go files that do not exist in database`
 	CGenDaoBriefTypeMapping       = `custom local type mapping for generated struct attributes relevant to fields of table`
 	CGenDaoBriefFieldMapping      = `custom local type mapping for generated struct attributes relevant to specific fields of table`
+	CGenDaoBriefWithPathPackage   = `is should make the dao package by the path`
```
gqcn, line 89: 这些修改与 https://github.com/gogf/gf/pull/3618 重复，建议从主分支新开分支单独修改`pgsql`的`RETURNING "ID"`。

gqcn (changes requested):

cmd/gf/internal/cmd/gendao/gendao.go
Insua: 用新的分支提交了，现在只有一个commit了
新pr是 #3638

PR #3691, opened 2024-07-19
fix(contrib/registry/file/v2): fix the panic caused by channel closing after `resolver closed`

zishang520:
This only fixes the problem of not triggering panic. As for why the resolver closed, I haven't investigated it yet and it will take more time.


zishang520:
@gqcn request review

contrib/registry/file/file_watcher.go
```diff
@@ -17,19 +18,26 @@ type Watcher struct {
...
-	return w.discovery.Search(context.Background(), gsvc.SearchInput{
-		Prefix: w.prefix,
-	})
+	if w.closed.Load() == nil {
+		<-w.ch
+		return w.discovery.Search(context.Background(), gsvc.SearchInput{
+			Prefix: w.prefix,
+		})
+	}
+	return nil, context.Canceled
```
gqcn, line 33: It might not appropriate using `context.Canceled` error here.
```go
if w.closed.Load() != nil {
    return nil, gerror.New("discovery service was closed")
}
<-w.ch
return w.discovery.Search(context.Background(), gsvc.SearchInput{
    Prefix: w.prefix,
})
```

contrib/registry/file/file_watcher.go
```diff
@@ -17,19 +18,26 @@ type Watcher struct {
...
 }
 
 // Close closes the watcher.
 func (w *Watcher) Close() error {
-	close(w.ch)
+	if w.closed.Load() == nil {
```
gqcn, line 38: It is appropriate using `*gtype.Bool` for attribute `w.closed`, which can use `Cas` to guarantee atomic operation.

gqcn (changes requested):

contrib/registry/file/file_watcher.go
zishang520: I looked at the context and this error is not used either:

https://github.com/gogf/gf/blob/9143b26182f49c6cb542e564721f89b3008ed92c/contrib/rpc/grpcx/internal/resolver/resolver_resolver.go#L58-L66

This will continue to run regardless of any abnormalities:

https://github.com/gogf/gf/blob/9143b26182f49c6cb542e564721f89b3008ed92c/contrib/rpc/grpcx/internal/resolver/resolver_resolver.go#L51-L67

So, I don't know what type of err to output.

contrib/registry/file/file_watcher.go
zishang520: Oh, I was not very familiar with the core of the framework before this PR. I will modify it later.

gqcn (approved):

PR #3693, opened 2024-07-19
fix(net/ghttp/): fix  the issue of using application/octet stream to upload large files causing excessive memory and crashing

TotoroZuo:
The reason for this problem is that when creating RouterFunc initializes the request, it calls the MakeBodyRepeatableRead method to write the r.body to memory and change it to Repeatable. However, pure binary files do not require formatting of either the form or the body

So the solution is to detect the application/octet stream type and exit it early in the formatted form and body methods

Fixes #3692

### DIFF
--- a/net/ghttp/ghttp_request_param.go
+++ b/net/ghttp/ghttp_request_param.go
@@ -229,6 +229,12 @@ func (r *Request) parseBody() {
 	if r.parsedBody {
 		return
 	}
+
+	// If it's binary data, it does not need to be parsed.
+	if contentType := r.Header.Get("Content-Type"); contentType == "" || gstr.Contains(contentType, "octet-stream") {
+		return
+	}
+
 	r.parsedBody = true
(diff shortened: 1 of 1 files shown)


net/ghttp/ghttp_request_param.go
```diff
@@ -229,6 +229,12 @@ func (r *Request) parseBody() {
 	if r.parsedBody {
 		return
 	}
+
+	// If it's binary data, it does not need to be parsed.
+	if contentType := r.Header.Get("Content-Type"); contentType == "" || gstr.Contains(contentType, "octet-stream") {
```
gqcn, line 234: - condition `contentType == ""` might be removed?
- missing associated unit testing case for the changes.

gqcn (changes requested):

gqcn:
@TotoroZuo Any updates?

Issue #3680, opened 2024-07-12
contrib/registry/nacos/v2: panic: duplicate metrics collector registration attempted


LanceAdd:
### Go version

go version go1.22.4 darwin/amd64

### GoFrame version

2.7.2

### Can this bug be reproduced with the latest release?

Option Yes

### What did you do?

> load remote config from nacos

```
serverConfig := constant.ServerConfig{
		IpAddr: "172.16.23.99",
		Port:   8848,
	}
	clientConfig := constant.ClientConfig{
		CacheDir:    "/tmp/nacos",
		LogDir:      "/tmp/nacos",
		NamespaceId: "c1e15245-03c1-4cba-9c40-ddc59e4d4167",
	}
	configParam := vo.ConfigParam{
		DataId: "sso.yml",
		Group:  "test",
	}

	adapter, err := nacos.New(ctx, nacos.Config{
		ServerConfigs: []constant.ServerConfig{serverConfig},
		ClientConfig:  clientConfig,
		ConfigParam:   configParam,
	})
	if err != nil {
		g.Log().Fatalf(ctx, `%+v`, err)
	}
	g.Cfg().SetAdapter(adapter)
```
> Register with Nacos
```
	gsvc.SetRegistry(nacos.New("172.16.23.99:8848",
		constant.WithNamespaceId("c1e15245-03c1-4cba-9c40-ddc59e4d4167"),
		constant.WithUsername("nacos"),
		constant.WithPassword("nacos"),
	).SetGroupName("test"))
```

> When I get the configuration file from Nacos and then register the service with Nacos, I get the following error
```
panic: duplicate metrics collector registration attempted

goroutine 1 [running]:
github.com/prometheus/client_golang/prometheus.(*Registry).MustRegister(0xc0000b9db0, {0xc0004a0300, 0x2, 0x2})
        /Users/shanyujie/go/pkg/mod/github.com/prometheus/client_golang@v1.12.2/prometheus/registry.go:403 +0x116
github.com/prometheus/client_golang/prometheus.MustRegister({0xc0004a0300, 0x2, 0x2})
        /Users/shanyujie/go/pkg/mod/github.com/prometheus/client_golang@v1.12.2/prometheus/registry.go:178 +0x3f
github.com/nacos-group/nacos-sdk-go/v2/common/monitor.init.0()
[...]

PR #3729, opened 2024-08-15
feat(contrib/registry/nacos/v2): Switch nacos-sdk-go dependencies to the official version

LanceAdd:
Switch nacos-sdk-go dependencies to the official version
Update the go.mod file and nacos related code to update the nacos-sdk-go dependency from the joy999 branch to the official version of nacos-group. This change includes updating the package path and referencing the official version of nacos-sdk-go/v2, as well as removing references to older versions. This change is to solve the startup failure issue caused by repeated initialization of monitor when using Nacos as the configuration center and registry center
fix: #3680

### DIFF
--- a/contrib/config/nacos/nacos.go
+++ b/contrib/config/nacos/nacos.go
@@ -9,6 +9,7 @@ package nacos
 
 import (
 	"context"
+	"sync"
 
 	"github.com/nacos-group/nacos-sdk-go/v2/clients"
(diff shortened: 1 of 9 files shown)


hailaz (approved):

gqcn:
@LanceAdd Thank you for your PR, but this is duplicated with https://github.com/gogf/gf/pull/3745

PR #3745, opened 2024-08-26
refactor(nacos-registry): use official nacos sdk instead of the third-party nacos sdk

lingcoder:
**Use the official nacos sdk instead of the third-party naocs sdk, and update the relevant library versions at the same time**


gqcn (approved):

gqcn:
https://github.com/gogf/gf/issues/3680

PR #3803, opened 2024-09-24
fix(database/gdb): invalid order by statement generated when multiple order inputs

LanceAdd:
```
dao.User.Ctx(ctx).Offset(10).Limit(1).Order("age asc", "name desc")
```
Bad SQL is generated
```
SELECT `id`, `name`, `age`
FROM `user`
ORDER BY `id` asc name desc
LIMIT 1 OFFSET 10
```
Generate the correct SQL statement
```
SELECT `id`, `name`, `age`
FROM `user`
ORDER BY `id` asc, `name` desc
LIMIT 1 OFFSET 10
```


gqcn:
@LanceAdd Would you please add associated unit testing case for this update?

LanceAdd:
> @LanceAdd Would you please add associated unit testing case for this update?

Of course, but in which unit test file should I add this unit test

contrib/drivers/mysql/mysql_z_unit_model_test.go
```diff
@@ -4786,3 +4787,22 @@ func Test_Model_FixGdbJoin(t *testing.T) {
 		t.Assert(gtest.DataContent(`fix_gdb_join_expect.sql`), sqlSlice[len(sqlSlice)-1])
 	})
 }
+
+func TestOrderByStatementGenerated(t *testing.T) {
```
gqcn, line 4791: `TestOrderByStatementGenerated` -> `Test_OrderBy_Statement_Generated`

contrib/drivers/mysql/mysql_z_unit_model_test.go
```diff
@@ -4786,3 +4787,22 @@ func Test_Model_FixGdbJoin(t *testing.T) {
...
+	gtest.C(t, func(t *gtest.T) {
+		array := gstr.SplitAndTrim(gtest.DataContent(`table_with_prefix.sql`), ";")
```
gqcn, line 4793: Please use standalone sql for your uint testing case, do not be coupled with other unit testing cases.

gqcn (changes requested):

contrib/drivers/mysql/mysql_z_unit_model_test.go
LanceAdd: OK

contrib/drivers/mysql/mysql_z_unit_model_test.go
LanceAdd: > Please use standalone sql for your uint testing case, do not be coupled with other unit testing cases.

OK

gqcn (approved):

gqcn:
@LanceAdd hello, please resolve the conflicts before merging.

gqcn (approved):

PR #3806, opened 2024-09-25
feat(os/gfsnotify):add watcher path to moniter the dir which create after start

TenYearOldJAVA:
When using the fsnotify library to monitor file system changes, a common requirement is to watch not only a specific directory but also all of its subdirectories, including those that are created after the monitoring process has started. However, fsnotify does not natively support recursive watching of directories. This limitation means that if a new subdirectory is created within a watched directory, changes within this new subdirectory will not be detected unless it is explicitly added to the watch list.

Manual Recursion: Users must manually implement a recursive mechanism to add all existing subdirectories to the watch list at the start and continue to monitor for new directories.So this function is append the watcher path，if path is not in callbackMap，it will not append.

### DIFF
--- a/os/gfsnotify/gfsnotify_watcher.go
+++ b/os/gfsnotify/gfsnotify_watcher.go
@@ -41,25 +41,21 @@ func (w *Watcher) Add(path string, callbackFunc func(event *Event), recursive ..
 // new sub-directories are captured, thus maintaining comprehensive monitoring coverage
 // without manual interventions after initial setup.
 func (w *Watcher) AppendPath(path string, recursive ...bool) error {
-
 	// Check and convert the given path to absolute path.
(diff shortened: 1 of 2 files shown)


gqcn:
@TenYearOldJAVA Thanks for contribution to goframe, we are reviewing this which might need some time.

gqcn:
@TenYearOldJAVA Hello, actually the gfsnotify package already implemented the recursive watching feature when  subfolders and sub-files created under folders that aleady watched, the codes are here: https://github.com/gogf/gf/blob/0ca2ed597971d860e18e00e78c5e4c9ef790a4e2/os/gfsnotify/gfsnotify_watcher_loop.go#L99

I also submit a new PR to improve the implements and add new unit testing case for this here: https://github.com/gogf/gf/pull/3830

PR #3841, opened 2024-10-07
feat(encoding/ghtml): add parameter validation for function `SpecialCharsMapOrStruct`

### LINKED ISSUE #3173: When html An error occurred when using input in SpecialCharsMapOrStruct
model
```
type AdminAddInput struct {
	Name     string  `json:"name"`     //用户名
	Password string  `json:"password"` //密码
	RoleIds  *string `json:"roleIds"`  //角色列表
	IsAdmin  uint8   `json:"isAdmin"`  //是否是管理员
}
```

logic
```
func (a *sAdmin) Add(ctx context.Context, input model.AdminAddInput) error {
	if err := ghtml.SpecialCharsMapOrStruct(input); err != nil {
		return gerror.New(exceptions.XssException)
	}
	//业务代码
        return nil
}
```

接口请求参数如下
The interface request parameters are as follows
image

报错信息如下
The error message is as follows
image

换成地址传值却正常调用
Change to address and pass value but call normally
image

我的另一个接口不使用地址传值却也正常运行
My other interface doesn't use address transfer, but it still runs normally
image
image

gqcn:
also fixed #3173


encoding/ghtml/ghtml.go
```diff
@@ -8,6 +8,8 @@
 package ghtml
 
 import (
+	"github.com/gogf/gf/v2/errors/gcode"
```
houseme, line 11: formatted import

oldme-git (approved):

encoding/ghtml/ghtml.go
gqcn: done

houseme (approved):

PR #3849, opened 2024-10-09
test(drivers/mysql): use asserts function to replace if statement

oldme-git:


gqcn (approved):

gqcn:
@oldme-git You are handsome and best 💖

PR #3896, opened 2024-10-28
feat(net/goai): add enhanced response status interface

### LINKED ISSUE #3889: net/goai: Operation object info improvement
### Description

@gqcn Currently GoFrame are using tags for openapi doc generation, most functions are fully functional and very helpful. But when adding some tags like `security`, there always needs to add a serialized json string, which is large and would makes the structure less readable. And too many tags would also have the same problem. Therefore, would it possible to add an interface for xxxReq to add a `HOOK` in the openapi doc generation process? Like we could add some functions here before using the generated `operation` object:

https://github.com/gogf/gf/blob/e4c4fb591e4f281c9406b7b01d0b24d703e1c9da/net/goai/goai_path.go#L278

``` go
if enhancedRequest, ok := outputObject.Interface().(ResponseStatusDef); ok {
    enhancedRequest.someHooker(&operation)
}
```

This could also cover some existing features that tags offers, but users could change the `operation` object directly in the hook function and override some info without limitations, making modifying openapi doc much easier and have more flexibility with some special needs. Also it could help reduce the tags info in the structure design that are not related to the real business info. And as the original design, the `goai` pack would provide 80% usage case, users could have this interface to change the doc in some edge cases.

While indeed, this feature has some duplication with what tags offers, but since it could be a much higher level operation, may be it could be added? Hoping to get your opinion for the framework perspective.

### Additional

_No response_

UncleChair:
Fixes #3889
- Add enhanced response interface for openapi config for users to add extra responses and examples
- Add related unit test
ok      github.com/gogf/gf/v2/net/goai  coverage: 80.2% of statements


gqcn:
@UncleChair Hello, thanks for your contribution, but I have a little suggestion for this implementing `OpenAPI Operation` enhancement:
- It might be better to add an interface like:
```go
type IEnhanceOperation interface{
    EnhanceOperation(operation *goai.Operation, req interface{}, res interface{})
}
```
- The previous defined interface `ResponseStatusDef` might be removed as it is not released officially.

gqcn:
@UncleChair 

> I have also considered to use interface solution to modify the `operation`, but I regard this feature to be a common method for users to add this to all api for processing some common settings, like what `common response` are doing. So it may be better to create a hook and users would not need to implement this interface everywhere?

Yes, it is reseanable. But the `goai.OpenApiV3` from `ghttp.Server.GetOpenApi()`already exposed its attributes public, which can be customized by users.  And also, the hook function is a little bit ugly, we may not add this hook feature for package `goai` util it is really necessary. But the `IEnhanceOperation` interface is good, we can add this as soon as possible.

gqcn:
@UncleChair Indeed, it would be quite complicated for users if only providing `IEnhanceOperation`. How about this:

1. We only keep the `ResponseStatusDef` design, which would be defined on `response` struct, but rename it to:
```go
// ResponseStatusStruct is the structure for certain response status.
type ResponseStatusStruct = any

// IEnhanceResponseStatus is used to enhance the documentation of the response.
// Normal response structure could implement this interface to provide more information.
type IEnhanceResponseStatus interface {
	EnhanceResponseStatus() map[StatusCode]ResponseStatusStruct
}
```

2. We do not provide `IEnhanceOperation` util it is necessary for some scenario.

UncleChair:
@gqcn Changed the interface related name
> 2. We do not provide `IEnhanceOperation` util it is necessary for some scenario.

Actually I am also tring to generate json files for examples during the `goai` doc generation process. In some cases I may already have an error list for a response with different status code, For example:

``` go
package v1

import (
	"cssgen-api/api"

	"github.com/gogf/gf/v2/errors/gcode"
)

// parameters should use uppercase
type LoginReq struct {
	g.Meta   `path:"/login" tags:"Auth" method:"post" sum:"User Login"`
	Username string `json:"username" v:"required-without:Email" des:"User Name" eg:"admin"`
	Email    string `json:"email" v:"required-without:Username" des:"User Email" eg:"admin@gmail.com"`
	Password string `json:"password" v:"required" des:"Password" eg:"adminPassword"`
}
type LoginRes struct {
	Token  string `json:"token" des:"JWT token"`
	Expire int64  `json:"expire" des:"JWT expire time"`
}

type LoginRes401 struct {
	g.Meta `status:"401" resEg:"resource/openapi/cssgen-api.api.auth.v1.LoginRes/401.json"`
}

func (r LoginRes) EnhanceResponseStatus() map[int]any {
	return map[int]any{
		401: LoginRes401{},
	}
}

var LoginErrRes = map[int][]gcode.Code{
	401: {
		gcode.New(1, "No related users", nil),
		gcode.New(2, "Wrong username or password", nil),
	}
}

func (r LoginRes) GenerateJsonFile() (path string, err error) {
	return api.GenerateJsonFile(r, LoginErrRes)
}
```

If I want to generate a static file for the error list, I have to use some mechanism to generate the file first and fill the generated file path `resource/openapi/cssgen-api.api.auth.v1.LoginRes/401.json` to the related response, which is not that automatic. This may be an edge case for example usage and we should desigen the api first, but the generation process may still important since we may need to sync some changes for error list if we modified it, which is also very common during project development, then I am trying to add a hook for operations in this pr.
I would like to provide some solutions for error list and example file synchronization in development, but since the structure of error list may differ from each other and we could not add a common interface for that, the best way seems to let users define what it should be and offer some hooks for users to`trigger` the process. Do you have any good idea to solve this usage problem?

gqcn:
> https://github.com/gogf/gf/pull/3896#issuecomment-2472172418
I see. It is quite complicated for users to handle `OpenAPIv3` structure, it is not easy and convenient.
How about expose the only predicted attributes that would be customized by users to a public structure, and then we add the attributes one by one to the structure if necessary. For example, we change the `EnhanceResponseStatus() map[int]any` to:
```go
type EnhancedStatusCode = int
type EnhancedStatusType struct{
    Response any
    Example   any
}
EnhanceResponseStatus() map[EnhancedStatusCode]EnhancedStatusType
```

Currently, we add `Example` to the `EnhancedStatusType` as it is necessary, we can easily extend it in the future if also necessary.

UncleChair:
@gqcn Would we keep the `ResponseStatusDef` related code since 2.8.0 was released?

gqcn (approved):

gqcn:
> @gqcn Would we keep the `ResponseStatusDef` related code since 2.8.0 was released?

We need not keep `ResponseStatusDef` as this feature is not truelly released as we do not mention it in the official document. We'll release this feature in next version.

PR #3897, opened 2024-10-29
Updates #3879: Handle the problem of inserting different libraries in a transaction

yimuysl001:
When compatible transactions insert data, the table structure of different libraries from the data source is abnormal, resulting in insertion failure. At present, only the places detected before have been processed, so there may be omissions


gqcn:
@yimuysl001 Hello, the ci failed, please have a check.

PR #3914, opened 2024-11-12
feat(net/goai): add min, max, length, min-length, max-length and between support for OpenAPIv3

ninjashixuan:
Improve OpenAPI Documentation Generation

This update primarily enhances the OpenAPI documentation generation by setting the min, max, and max-length constraints for the schema based on request validation.


net/goai/goai_shema.go
```diff
@@ -204,6 +205,61 @@ func (oai *OpenApiV3) structToSchema(object interface{}) (*Schema, error) {
 			if validationRuleSet.Contains(validationRuleKeyForRequired) {
 				schema.Required = append(schema.Required, key)
 			}
+
+			// Extract validation rules to schema. like min, max, length
+			validationRules := ref.Value.ValidationRules
+			// remove custom error message
+			if idx := strings.Index(ref.Value.ValidationRules, "#"); idx != -1 {
```
gqcn, line 212: You can refer current parsing logic in package `goai`, which uses `gvalid.ParseTagValue` to parse the validation pattern: https://github.com/gogf/gf/blob/e8a2629b19ddbd4144bbd69533ef4e6dd00ace5e/net/goai/goai_shema.go#L230

net/goai/goai_shema.go
```diff
@@ -204,6 +205,61 @@ func (oai *OpenApiV3) structToSchema(object interface{}) (*Schema, error) {
...
+				validationRules = ref.Value.ValidationRules[:idx]
+			}
+			lstRules := gstr.Split(validationRules, "|")
+			for _, rule := range lstRules {
+				if strings.HasPrefix(rule, "max") {
```
gqcn, line 217: Do not use hard-coded string, use constant instead. Please refer current usage example: https://github.com/gogf/gf/blob/7d4c59ac5aea64a6db83e79d6bb7aee1ed66250e/net/goai/goai.go#L66

gqcn (changes requested):

gqcn:
@ninjashixuan This PR is awsome! I'm watching this.

net/goai/goai_shema.go
ninjashixuan: up to this point, gvalid.ParseTagValue(validationTagValue) has already been processed. The rules have already removed the custom error messages, and strings.index is no longer needed.

net/goai/goai_shema.go
ninjashixuan: Indeed, hard-coded is not good practice.

net/goai/goai.go
```diff
@@ -65,6 +65,12 @@ const (
 const (
 	validationRuleKeyForRequired = `required`
 	validationRuleKeyForIn       = `in:`
+	validationRuleKeyMax         = `max`
```
gqcn, line 68: - Use `max:` instead of `max` for following string prefix check. And the same as `min:`, `length:`, `max-length:`, etc.
- Missing unit testing cases for this updtae.

gqcn (changes requested):

net/goai/goai_shema.go
gqcn: it is indeed.

PR #3925, opened 2024-11-14
feat(contrib/drivers): add gaussdb driver

okyer:


contrib/drivers/gaussdb/gaussdb.go
```diff
@@ -0,0 +1,54 @@
+// Copyright GoFrame Author(https://goframe.org). All Rights Reserved.
```
gqcn, line 1: @okyer Hello, I see the CI of this PR failed, as the docker service is missing for unit testing cases of gaussdb. You can add your gaussdb docker service in CI yaml here https://github.com/gogf/gf/blob/master/.github/workflows/ci-main.yml . You can refer to the existing docker service in CI yaml.

gqcn (changes requested):

gqcn:
https://github.com/gogf/gf/pull/4563

Issue #3930, opened 2024-11-16
net/goai: embedded struct attributes in response struct do not show in OpenAPIv3 document


gqcn:
### Go version

1.23.1

### GoFrame version

v2.8.0-beta

### Can this bug be reproduced with the latest release?

Option Yes

### What did you do?

In quick start api document, the `Get one user` api's response attributes do not show in OpenAPIv3 document.

Codes source repo: https://github.com/gogf/quick-demo

![QQ_1731751422780](https://github.com/user-attachments/assets/45e67a8c-472f-49b0-a540-5e0a1eebaa30)

![QQ_1731751299982](https://github.com/user-attachments/assets/fd514bdc-ba77-4785-b5b4-4bedd721acf4)

### What did you see happen?

The `Get one user` api's response attributes do not show in OpenAPIv3 document.

### What did you expect to see?

The `Get one user` api's response attributes should show in OpenAPIv3 document.

PR #3956, opened 2024-11-21
fix(net/goai): embedded struct with tags also expand as properties of the parent struct

wlynxg:
fix: https://github.com/gogf/gf/issues/3930


gqcn:
@wlynxg Please submit associated unit testing case for this update.

gqcn (approved):

PR #4000, opened 2024-12-02
feat(contrib/drivers/pgsql): add array type `varchar[]` and `text[]` converting to Go `[]string` support

ninjashixuan:


gqcn:
@ninjashixuan Please add associated unit testing case for your commit in package `contrib/drivers/pgsql`.

ninjashixuan:
@gqcn Oops, I forgot to add the test cases earlier. they're in now.

gqcn (approved):

ninjashixuan:
[https://github.com/ninjashixuan/gf/blob/baad29c1cc1774762a4a0e8ecc94a18f7175efb3/contrib/drivers/pgsql/pgsql_convert.go#L124-L130](url)
This shouldn't return the pg.StringArray. Instead, return []string(result). This way, gconv.Strings skips an extra layer of reflection and assignment， reducing some performance overhead. I won’t submit a separate PR for this minor change.
@gqcn

PR #4046, opened 2024-12-14
perf(contrib/drivers/pgsql): improve conversion performace for slice string

ninjashixuan:
enhance #4000，One line change, Return basic types instead of pq.StringArray to reduce one reflection assignment when calling gconv.Strings in doStruct, minimizing performance overhead.

### DIFF
--- a/contrib/drivers/pgsql/pgsql_convert.go
+++ b/contrib/drivers/pgsql/pgsql_convert.go
@@ -128,7 +128,7 @@ func (d *Driver) ConvertValueForLocal(ctx context.Context, fieldType string, fie
 		if err := result.Scan(fieldValue); err != nil {
 			return nil, err
 		}
-		return result, nil
+		return []string(result), nil
 
 	default:
 		return d.Core.ConvertValueForLocal(ctx, fieldType, fieldValue)


contrib/drivers/pgsql/pgsql_convert.go
```diff
@@ -128,7 +128,7 @@ func (d *Driver) ConvertValueForLocal(ctx context.Context, fieldType string, fie
 		if err := result.Scan(fieldValue); err != nil {
 			return nil, err
 		}
-		return result, nil
+		return []string(result), nil
```
gqcn, line 131: The `pq.StringArray` is type of `[]string`, the `[]string(result)` is `[][]string`, you sure?

gqcn (changes requested):

contrib/drivers/pgsql/pgsql_convert.go
ninjashixuan: Sure, []string(result) is only converting pq.ArrayString to the base type []string, not [][]string. As a result, when calling gconv.Strings, it directly matches the case []string branch instead of relying on originValueAndKind := reflection.OriginValueAndKind(any).

contrib/drivers/pgsql/pgsql_convert.go
```diff
@@ -128,7 +128,7 @@ func (d *Driver) ConvertValueForLocal(ctx context.Context, fieldType string, fie
 		if err := result.Scan(fieldValue); err != nil {
```
gqcn, line 128: I found it here should be `result.Scan(&fieldValue)`.

For example:
```go
package main

import (
	"fmt"

	"github.com/lib/pq"
)

func main() {
	var s pq.StringArray
	err := gconv.Scan([]string{"a", "b"}, &s)
	fmt.Println(err)
	fmt.Println(s)
}
```

contrib/drivers/pgsql/pgsql_convert.go
gqcn: OK, I see.

gqcn (changes requested):

contrib/drivers/pgsql/pgsql_convert.go
ninjashixuan: I don't think so.  fieldValue is of type []byte. The pg.StringArray Scan method only accepts []byte, string, or nil, and using &fieldValue as *[]byte would result in a conversion error.

gqcn (approved):

PR #4059, opened 2024-12-18
fix(database/gdb): regular expression pattern for link configuration to be compitable with tidbcloud

robotism:
---
tidb cloud link is like

```
tidb:2hcmRccccxxx9Fizz.root:wP3xxxxPIDc@tcp(xxxx.tidbcloud.com:4000)/db_name?tls=true
```
username has a dot `2hcmRccccxxx9Fizz.root`

so need update the pattern to `(\w+):([^:]*):(.*?)@(\w+?)\((.+?)\)/{0,1}([^\?]*)\?{0,1}(.*)`

---
the origin pattern is
linkPattern = `(\w+):([\w\-\$]*):(.*?)@(\w+?)\((.+?)\)/{0,1}([^\?]*)\?{0,1}(.*) `
 it is uncompitable with tidb cloud link;

of cause, i can use
```
- type: "tidb"
  host: "xxxxx
  port: "4000"
  user: "2hcmRccccxxx9Fizz.root"
  pass: "wP3xxxxPIDc"
  name: "db_name"
```
to instead of the link parameter, it works;

### DIFF
--- a/.github/workflows/ci-main.sh
+++ b/.github/workflows/ci-main.sh
@@ -1,8 +1,5 @@
 #!/usr/bin/env bash
 
-# Define the latest Go version requirement
-LATEST_GO_VERSION="1.23"
-
 coverage=$1
 
 # find all path that contains go.mod.
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -28,10 +28,21 @@ concurrency:
 
 env:
   TZ: "Asia/Shanghai"
-
+  # for unit testing cases of some components that only execute on the latest go version.
+  LATEST_GO_VERSION: "1.23"
 
 jobs:
   code-test:
+    strategy:
+      matrix:
+        # 🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥
+        # When adding new go version to the list, make sure:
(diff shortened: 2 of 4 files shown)


database/gdb/gdb.go
```diff
@@ -708,7 +708,7 @@ const (
 	ctxKeyCatchSQL            gctx.StrKey = `CtxKeyCatchSQL`
 	ctxKeyInternalProducedSQL gctx.StrKey = `CtxKeyInternalProducedSQL`
 
-	linkPattern            = `(\w+):([\w\-\$]*):(.*?)@(\w+?)\((.+?)\)/{0,1}([^\?]*)\?{0,1}(.*)`
+	linkPattern            = `(\w+):([^:]*):(.*?)@(\w+?)\((.+?)\)/{0,1}([^\?]*)\?{0,1}(.*)`
```
gqcn, line 711: - It would be better using `^(\w+):(.*?):(.*?)@(\w+?)\((.+?)\)/{0,1}([^\?]*)\?{0,1}(.*?)$` for the pattern.
- Necessary unit testing case should be added for this update here: https://github.com/gogf/gf/blob/92eab81926dea3ef090905c8f4da267c84fd11a3/database/gdb/gdb_z_mysql_internal_test.go#L297

gqcn (changes requested):

database/gdb/gdb.go
robotism: ok, i open a new pr #4064

Issue #4078, opened 2024-12-25
ghttp: Can support setting default values ​​for header?


nilsir:
### Is your feature request related to a problem?

Option Yes

### Describe the solution you'd like

I want to set a default value when the header information is empty

### Describe alternatives you've considered

```golang
# now is 
language := r.GetHeader("lang")
if language == "" {
    language = "zh-CN"
}

# replace is
r.GetHeader("lang", "zh-CN") 
```

gqcn:
Good proposal!

PR #4090, opened 2024-12-31
feat(net/ghttp): add default value support ​​for retrieving header value

sanrentai:
feat(net/ghttp): support setting default values ​​for header #4078

### DIFF
--- a/util/gconv/gconv_z_unit_scan_test.go
+++ b/util/gconv/gconv_z_unit_scan_test.go
@@ -126,7 +126,7 @@ func TestScan(t *testing.T) {
 			err = gconv.Scan(maps, &mss)
 			t.AssertNil(err)
 			t.Assert(len(mss), len(maps))
-			for k, _ := range maps {
+			for k := range maps {
 				t.Assert(maps[k]["Name"], mss[k]["Name"])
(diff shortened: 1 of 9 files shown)


net/ghttp/ghttp_request.go
```diff
@@ -151,8 +151,14 @@ func (r *Request) IsExited() bool {
 }
 
 // GetHeader retrieves and returns the header value with given `key`.
-func (r *Request) GetHeader(key string) string {
-	return r.Header.Get(key)
+func (r *Request) GetHeader(key string, def ...string) string {
```
gqcn, line 154: Please add associated unit testing case for this update.

contrib/drivers/mssql/mssql_open.go
```diff
@@ -30,6 +34,13 @@ func (d *Driver) Open(config *gdb.ConfigNode) (db *sql.DB, err error) {
 		)
 		return nil, err
 	}
+	if err = db.QueryRow("SELECT @@VERSION").Scan(&version); err != nil {
```
gqcn, line 37: Why adding this update in this PR?

gqcn (changes requested):

Issue #4104, opened 2025-01-08
os/gtime: issue title polaris configuration center, only supports json data and does not support other file types (such as: yaml)


sharpleafage:
### Go version

go.1.23

### GoFrame version

gf >=2.8.0

### Can this bug be reproduced with the latest release?

Option Yes

### What did you do?

dev

### What did you see happen?

加载配置中心的问题，
gf >=2.8.0  

```
func (c *Client) doUpdate(ctx context.Context) (err error) {
	if !c.client.HasContent() {
		return gerror.New("config file is empty")
	}
	var j *gjson.Json
	if j, err = gjson.DecodeToJson([]byte(c.client.GetContent())); err != nil {
		return gerror.Wrap(err, `parse config map item from polaris failed`)
	}
	c.value.Set(j)
	return nil
}
```

gf <2.8.0  
```
func (c *Client) doUpdate(ctx context.Context) (err error) {
	if !c.client.HasContent() {
		return gerror.New("config file is empty")
	}
	var j *gjson.Json
	if j, err = gjson.LoadContent(c.client.GetContent()); err != nil {
		return gerror.Wrap(err, `parse config map item from polaris failed`)
	}
	c.value.Set(j)
	return nil
}

```
```

//bug
 gjson.DecodeToJson([]byte(c.client.GetContent())); 

```

### What did you expect to see?

解决bug

sharpleafage:
按照文档 ，代码行不通， gf2.7版本，可以行的通
https://goframe.org/docs/micro-service/config-service

stardemo:
遇到了相同的问题

lingcoder:
polaris 这边后来改回 gjson.LoadContent 了，yaml 这些格式可以自动识别，≥v2.9.0。
这单关了。

PR #4126, opened 2025-01-20
fix(contrib/config/polaris): it only supports json format for configuration contents, add more content formats support

stardemo:
polaris conf update use gjson.LoadContent not gjson.DecodeToJson to support multiple format conf content.

Fixs #4104 4104


gqcn:
@stardemo Awesome fix! Would you please add associated unit testing case for this update?

stardemo:
this unit testing may difficult to create,but this change work fine in my local environment.

gqcn (approved):

PR #4182, opened 2025-03-04
fix(net/ghttp):check parameter existence to determine using default or front-end value.

### LINKED ISSUE #4093: os/gtime: issue title
### Go version

arm64

### GoFrame version

ex 2.7

Option Yes

### What did you do?

{
    "state": "",
    "pagination": false
} 请求参数中的false会被解析为true,只有0才会false

### What did you see happen?

{
    "state": "",
    "pagination": false
} 请求参数中的false会被解析为true,只有0才会false

{
    "state": "",
    "pagination": false
} 请求参数中的false会被解析为true,只有0才会false

river291:
- Fixes #4093 
The results of frontend-passed values are ultimately converted into key-value pairs. When a key exists, the value has only two possible states: present or absent.

For non-existent keys, there are two scenarios:
1. The value might inherently not exist for that key.
2. A default value (e.g., from the default/d tag) is provided.

Implementation logic:
 - If the key does NOT exist: Use the default value.
 - If the key EXISTS:
  1. Value is absent: Fall back to the default value.
  2. Value is present: Directly use the frontend-provided value without additional checks (e.g., no IsEmpty() validation as in previous versions).

### DIFF
--- a/net/ghttp/ghttp_request_param_request.go
+++ b/net/ghttp/ghttp_request_param_request.go
@@ -204,8 +204,11 @@ func (r *Request) mergeDefaultStructValue(data map[string]interface{}, pointer i
 				if foundKey == "" {
 					data[field.Name()] = tagValue
 				} else {
-					if empty.IsEmpty(foundValue) {
+					// Check parameter existence to determine using default or front-end value.
+					if foundValue == nil || foundValue == "" {
 						data[foundKey] = tagValue
+					} else {
+						data[foundKey] = foundValue
 					}
 				}
 			}
@@ -228,8 +231,11 @@ func (r *Request) mergeDefaultStructValue(data map[string]interface{}, pointer i
 			if foundKey == "" {
 				data[field.Name()] = field.TagValue
 			} else {
-				if empty.IsEmpty(foundValue) {
+				// Check parameter existence to determine using default or front-end value.
+				if foundValue == nil || foundValue == "" {
 					data[foundKey] = field.TagValue
+				} else {
+					data[foundKey] = foundValue
 				}
 			}
 		}


wln32 (approved):

net/ghttp/ghttp_request_param_request.go
```diff
@@ -261,34 +240,31 @@ func (r *Request) mergeInTagStructValue(data map[string]interface{}) error {
...
 			}
 		}
 	}
 	return nil
 }
+
+// mergeFieldValue merges the request parameters when the key does not exist in the map or isCover is true or the value is nil.
+//
+// createTime: 2025-03-05 17:50:32
+func mergeFieldValue(data map[string]interface{}, isCover bool, findKey string, fieldName string, tagValue interface{}) {
```
gqcn, line 262: 1. `mergeFieldValue`的方法名称比较宽泛，不满足"见名知意"的命名习惯，建议修改为`mergeTagValueWithFoundKey`。
1. 去掉 `createTime: 2025-03-05 17:50:32` 这段注释，通过`git blame`可以看得到修改时间和记录。
2. `isCover`变量命名不太恰当，按照我对这个变量的理解，建议修改为`overwritten`。
3. 建议在`issue`的单例文件中 https://github.com/gogf/gf/blob/master/net/ghttp/ghttp_z_unit_issue_test.go 增加对该pr修复的特定issue https://github.com/gogf/gf/issues/4093 的单元测试覆盖，以验证修改符合预期。单例编写方式请参考已有的单例代码。

gqcn (changes requested):

gqcn (approved):

PR #4220, opened 2025-03-24
feat(util/gconv): support automatic conversion of `Y/N` string to bool

FSpark:
In Oracle databases, it is very common to use CHAR(1) fields with 'Y' or 'N' to represent boolean values. This PR adds support for interpreting 'Y' as true and 'N' as false to improve compatibility and developer experience.

Note:
Should we also consider adding support for 'T'/'F' as boolean equivalents? This pattern is also occasionally used in some legacy systems.


gqcn:
@FSpark You still updated the global converter. Such updates should be made in oracle implements.

gqcn:
closed due to incativity

Issue #4227, opened 2025-03-28
net/ghttp: Param in header would affect the following params parse


UncleChair:
### Go version

go version go1.23.2 windows/amd64

### GoFrame version

2.9.0

### Can this bug be reproduced with the latest release?

Option Yes

### What did you do?

Initial web project with the latest version and modify the `HelloReq` to :

```
type HelloReq struct {
  g.Meta  `path:"/hello" tags:"Project" method:"get" sum:"List all projects"`
  Token   string `json:"Authorization" des:"JWT" in:"header"`
  Owned   bool   `json:"owned" in:"query" default:"false"`
  Deleted bool   `json:"deleted" in:"query" default:"false"`
}
```

Send the request and dump req in controller

```
curl --location 'http://localhost:8000/hello?deleted=false' \
--header 'Authorization: 1'
```

### What did you see happen?

The dump info would show:

```
{
    Token:   "1",
    Owned:   true,
    Deleted: false,
}
```

### What did you expect to see?

In the previous version (2.8.3), it was:
```
{
    Token:   "1",
    Owned:   false,
    Deleted: false,
}
```

Or if you move the Token to the end of the structure, it would also have the right dump:
```
{
    Owned:   false,
    Deleted: false,
    Token:   "1",
}
```

Seems some changes was made with the param parse? Haven't locate the bug yet, may be gconv related

UncleChair:
https://github.com/gogf/gf/blob/fee38b4531224b929c095ef643a68a92bbad3d4b/net/ghttp/ghttp_request_param_request.go#L241-L254

@gqcn Missing in query check here, seems related to #4182 ?

PR #4228, opened 2025-03-28
fix(net/ghttp): wrong in-tag param parse for query param (#4227)

UncleChair:
Fixed issue #4227 and add unit test


UncleChair:
@gqcn Any update sir? This problem could be serious if any logic is relied on the parsing.

UncleChair:
@hailaz @houseme Request for a review, thx

PR #4324, opened 2025-06-24
feat(util/gutil): append finally func parameter to Try/TryCatch/Go

CharLemAznable:


gqcn:
@CharLemAznable To better convey the function's meaning, why not add a new `TryCatchFinal` function instead?

CharLemAznable:
> @CharLemAznable To better convey the function's meaning, why not add a new `TryCatchFinal` function instead?

@gqcn  First, if a new function `TryCatchFinal` is added to complement the existing `TryCatch` function, then a corresponding `TryFinal` function should also be added to complement the `Try` function. Additionally, new methods may need to be added to `gutil_groutine` and `g_func`. In this scenario, there would be a significant amount of redundant code, increasing the burden of future modifications and maintenance.

Another reason is that the optional `finally` block is a common part of `try-catch` syntax. If added as an optional parameter to the `TryCatch` and `Try` methods, it would not impose a large cognitive burden on most users. Compared to adding optional parameters, introducing new methods represents a more significant API change.

gqcn:
I cannot complete this review reliably yet, so it would help to have a maintainer take a look.

The reason is: `AGENTS.md` is not present on the PR's target commit, so the code cannot be checked against the repository rules from that base.

Suggested focus: Try/TryCatch/Go finally parameter

Suggested reviewers, if available: @huangqian1985 @houseme @LonelySally

Why they are mentioned: they have worked on related files.

PR #4334, opened 2025-07-07
fix(tracing): set database span kind to client

minimAluminiumalism:
In [OpenTelemetry spec](https://opentelemetry.io/docs/specs/semconv/database/database-spans/), the span kind of database should be `client`.


minimAluminiumalism:
@gqcn Could you please take a look when you have a moment?

PR #4441, opened 2025-09-22
fix(cmd/gf): run AddSigHandlerShutdown cannot work well)

wanghaolong613:
https://github.com/gogf/gf/issues/3752

Sending [Interrupt] on Windows is not implemented.


wanghaolong613:
waitCtx, cancel := context.WithTimeout(ctx, 30*time.Second) 我加了ctx控制协程退出

PR #4452, opened 2025-09-28
fix(database/gdb): Fix GetArray return type and add Bools method

hailaz:
Change the return type of the GetArray method to Array for consistency in data structure. Introduce a new Bools method to convert Vars to a slice of bools, along with corresponding test cases to validate the functionality.

### DIFF
--- a/frame/gins/gins_database.go
+++ b/frame/gins/gins_database.go
@@ -89,7 +89,7 @@ func Database(name ...string) gdb.DB {
 				}
 			}
 			if len(cg) > 0 {
-				if gdb.GetConfig(group) == nil {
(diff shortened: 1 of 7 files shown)


frame/gins/gins_database.go
```diff
@@ -89,7 +89,7 @@ func Database(name ...string) gdb.DB {
 				}
 			}
 			if len(cg) > 0 {
-				if gdb.GetConfig(group) == nil {
+				if gcg, _ := gdb.GetConfigGroup(group); gcg == nil {
```
houseme, line 92: 返回的第二个参数是`err`吗？

hailaz: 对，这里的修改是因为lint报错了，逻辑和原来保持一致。

PR #4482, opened 2025-10-17
feat(container/garray): enhance generic array implements

joy999:
Remove the t array of wrapper array. Now it's a real one. Other normal array will base on it.


container/garray/garray_normal_str.go
```diff
@@ -725,32 +424,25 @@ func (a *StrArray) Iterator(f func(k int, v string) bool) {
...
 // String returns current array as a string, which implements like json.Marshal does.
 func (a *StrArray) String() string {
 	if a == nil {
 		return ""
 	}
+
+	a.lazyInit()
+
 	a.mu.RLock()
 	defer a.mu.RUnlock()
```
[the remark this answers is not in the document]
joy999: improved

container/garray/garray_z_unit_normal_t_test.go
```diff
@@ -81,14 +84,16 @@ func Test_TArray_Basic(t *testing.T) {
 		t.Assert(ok, true)
 
 		t.Assert(array.Contains(100), false)
-		array.Append(4) // 1, 2, 3 ,4
+		array.Append(4) // 2, 2, 3, 4
+		t.Assert(array.Slice(), []int{2, 2, 3, 4})
```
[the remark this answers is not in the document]
joy999: No need. It's correct.

container/garray/garray_normal_int.go
```diff
@@ -7,25 +7,17 @@
...
-	"github.com/gogf/gf/v2/util/grand"
 )
 
 // IntArray is a golang int array with rich features.
 // It contains a concurrent-safe/unsafe switch, which should be set
 // when its initialization and cannot be changed then.
 type IntArray struct {
-	mu    rwmutex.RWMutex
-	array []int
+	*TArray[int]
```
gqcn, line 21: Why use pointer here instead of value one?

container/garray/garray_normal_int.go
```diff
@@ -76,78 +66,64 @@ func NewIntArrayFrom(array []int, safe ...bool) *IntArray {
...
 	newArray := make([]int, len(array))
 	copy(newArray, array)
-	return &IntArray{
-		mu:    rwmutex.Create(safe...),
-		array: newArray,
+	return NewIntArrayFrom(newArray, safe...)
+}
+
+// lazyInit lazily initializes the array.
+func (a *IntArray) lazyInit() {
```
gqcn, line 75: Is it possible remove `lazy initialization` design?

container/garray/garray_z_unit_normal_t_test.go
```diff
@@ -608,7 +613,7 @@ func TestTArray_RLockFunc(t *testing.T) {
 
 		// 防止ci抖动,以豪秒为单位
 		t.AssertLT(t2-t1, 20) // go1加的读锁，所go2读的时候，并没有阻塞。
```
gqcn: Please use English instead of Chinese comments in source codes.

container/garray/garray_normal_int.go
joy999: There will be a warning when copy mu with value type. Use pointer will be more safer.

container/garray/garray_z_unit_normal_t_test.go
joy999: resolved

container/garray/garray_normal_int.go
joy999: Disscused with haiza. Maybe we need keep pointer and lazy init. Because there will be a warning when the lock copy by value type. We can discuss in qq group.

joy999: The important problem is "var a IntArray". It won't call new. So we need lazyInit.

PR #4485, opened 2025-10-18
refactor(database/gdb): add quote for FieldsPrefix

hicolin:
Code example：
``` go
	var res *BasicInfo
	err := g.DB().Model("basic_info").
		FieldsPrefix("basic_info", basicInfoColumns).
		Where("id", 35813305356386305).Scan(&res)
	if err != nil {
		panic(err)
	}
	g.Dump(res)
```

SQL generated before modification ：
``` sql
SELECT basic_info.id,basic_info.full_name,basic_info.contact FROM `basic_info` WHERE (`id`=35813305356386305) AND `delete_time` IS NULL LIMIT 1
```

SQL generated after modification：
``` sql
SELECT `basic_info`.`id`,`basic_info`.`full_name`,`basic_info`.`contact` FROM `basic_info` WHERE (`id`=35813305356386305) AND `delete_time` IS NULL LIMIT 1
```

### DIFF
--- a/database/gdb/gdb_model_fields.go
+++ b/database/gdb/gdb_model_fields.go
@@ -45,8 +45,9 @@ func (m *Model) FieldsPrefix(prefixOrAlias string, fieldNamesOrMapStruct ...any)
 	if len(fields) == 0 {
 		return m
 	}
+	prefixOrAlias = m.QuoteWord(prefixOrAlias)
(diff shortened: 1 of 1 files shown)


gqcn:
@hicolin The CI failed, you need update the expect result of the unit testing case for the CLI, too.

PR #4375, opened 2025-08-19
fix:修复pgsql多schema模式下相同表名中字段叠加的问题

iamcc30:

PR #4491, opened 2025-10-22
fix(contrib/drivers/pgsql): ensure fields queries are scoped to the specified schema

qinyuguang:
This PR addresses an issue where database metadata queries, specifically those retrieving table and column information using PostgreSQL catalog views (`pg_attribute`, `pg_class`, etc.), were not correctly scoped by schema.

**Problem:**
When multiple schemas contain tables with the same name (e.g., `schema1.some_table` and `public.some_table`), the previous query would return metadata for all tables, leading to inaccurate or duplicate results.

**Solution:**
The query logic has been updated to explicitly join with `pg_namespace` and filter results based on the schema name (`nspname`).

1.  Joined `pg_class` with `pg_namespace` via `c.relnamespace = n.oid`.
2.  Added a `WHERE` clause condition: `n.nspname = 'public'` (or the target schema name).
3.  Ensured the join to `information_schema.columns` is also filtered by `ic.table_schema = n.nspname` for accurate data retrieval (e.g., `default_value`, `length`).

This ensures that the metadata query correctly retrieves information only for the table within the intended schema.

### DIFF
--- a/contrib/drivers/pgsql/pgsql_table_fields.go
+++ b/contrib/drivers/pgsql/pgsql_table_fields.go
@@ -27,7 +27,8 @@ FROM pg_attribute a
          left join pg_description b ON a.attrelid=b.objoid AND a.attnum = b.objsubid
          left join pg_type t ON a.atttypid = t.oid
          left join information_schema.columns ic on ic.column_name = a.attname and ic.table_name = c.relname
-WHERE c.relname = '%s' and a.attisdropped is false and a.attnum > 0
+         left join pg_namespace n on c.relnamespace = n.oid
+WHERE c.relname = '%s' and n.nspname = '%s' and a.attisdropped is false and a.attnum > 0
 ORDER BY a.attnum`
 )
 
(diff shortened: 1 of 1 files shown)


qinyuguang:
This bug has a direct impact on our database generation tools, specifically when using `gf gen dao`.

When `gf gen dao` is executed, it incorrectly merges the column definitions from identically named tables across multiple schemas into a single struct. This results in the generated Go struct containing combined, non-existent, or duplicated field names.

Consequently, when this generated dao is used for gdb `Scan` operations, the SQL query attempts to select fields that are not present in the intended single table, leading to an immediate SQL error (e.g., "column does not exist") or incorrect query results.

gqcn:
@qinyuguang Hello, the CI failed, you might also update the source codes of the CLI.

qinyuguang:
I see. The "schema" in the parameters is not the actual schema in PostgreSQL. I'll replace it with "namespace" to make the distinction. 
The naming of the variables is a bit confusing.

hailaz:
和这个pr有没有关联？https://github.com/gogf/gf/pull/4375

qinyuguang:
> 和这个pr有没有关联？#4375

对是这个问题，用regclass很方便，我关掉这个PR

PR #4521, opened 2025-11-23
fix(database/gdb): Resolve column ambiguity in GROUP BY/ORDER BY with MySQL JOIN

ljluestc:
When using JOIN queries in MySQL with the `Group()` method, column names in GROUP BY clauses become ambiguous if multiple tables contain columns with the same name (commonly `id`). This results in MySQL errors like "Column 'id' in group statement is ambiguous".

**Example Issue:**
```go
model := t.Ctx(ctx).Fields("t_inf_job.*, t_inf_job_attr.*").
    LeftJoin("t_inf_job_attr", "t_inf_job.id = t_inf_job_attr.job_id").
    Where(t.Columns().Deleted, 0)

// This would fail with "Column 'id' in group statement is ambiguous"
err = model.Group(t.Columns().Id).Scan(&jobs)
```

### **Key Changes**

1. **Modified function signature**: `Group(groupBy ...string)` → `Group(groupBy ...any)` to support Raw SQL expressions
2. **Auto-prefixing logic**: When JOINs are detected (by checking for " JOIN " in the tables string), unqualified column names are automatically prefixed with the primary table name
3. **Preserved existing behavior**: Already qualified columns (containing ".") and Raw expressions are handled as before
4. **Added comprehensive test**: `Test_Model_Group_WithJoin` verifies the fix works correctly with JOIN queries

### DIFF
--- a/database/gdb/gdb_model_select.go
+++ b/database/gdb/gdb_model_select.go
@@ -749,34 +749,41 @@ func (m *Model) getHolderAndArgsAsSubModel(ctx context.Context) (holder string, 
 	return
 }
 
+func (m *Model) parseTableAlias(tableInitStr string) (alias string, tableName string) {
+	// Split by space to separate table from alias
+	// Format can be: `table` alias or `table` AS `alias` or just table alias
+	parts := gstr.SplitAndTrim(tableInitStr, " ")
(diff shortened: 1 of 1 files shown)


hailaz:
撤销了本次修改，详情 #4555

PR #4526, opened 2025-11-26
feat(contrib/drivers/opengauss): add driver support for openGauss

aaronsmile:
Add driver support for openGauss. The primary task is to replace `libpq` with the adapted `openGauss-connector-go-pq`.

When using original driver to access openGauss, the following error will be reported.
`pq: Invalid username/password,login denied.`
The reason is that openGauss default user connection password authentication method is sha256, which is a unique encryption method, the modified `libpq` can be directly compatible with sha256.

### DIFF
--- a/contrib/drivers/opengauss/opengauss_z_unit_init_test.go
+++ b/contrib/drivers/opengauss/opengauss_z_unit_init_test.go
@@ -20,12 +20,12 @@ import (
 )
 
 const (
-	TableSize          = 10
-	TablePrefix        = "t_"
-	SchemaName         = "test"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-main.yml
```diff
@@ -221,6 +236,13 @@ jobs:
         detached: true
         limit-access-to-actor: false
 
+    - name: Free Disk Space
```
hailaz, line 239: 好东西，直接改用你的方法清理空间

aaronsmile: thanks!

PR #4555, opened 2025-12-09
refactor(database/gdb): simplify order and group by alias quoting (bu…

### LINKED ISSUE #4554: bug(database/gdb): v2.9.6 Order() incorrectly adds original table name prefix instead of alias when using As()
### Go version

 go 1.25+

### GoFrame version

 v2.9.6

Option Yes

### What did you do?

 When using `.As("a")` to set a table alias with `LeftJoin`, and then using `Order(dao.Table.Columns().Field)`:

  ```go
  stickerCols := dao.DemoA.Columns()

  err = dao.DemoA.Ctx(ctx).As("a").
      LeftJoin(dao.DemoB.Table(), "b", "a.id=b.data_id").
      Fields("a.id", "a.name").
      Order(dao.DemoA.Columns().Sort).
      Scan(&res)

  What did you expect to see?

  v2.9.5 generates (correct):
  SELECT a.id, a.name
  FROM "demo_a" AS a
  LEFT JOIN "demo_b" AS "b" ON (a.id=b.data_id)
  ORDER BY "sort"
  LIMIT 10 OFFSET 0

  The Order() should generate ORDER BY "sort" (column name only), as it did in v2.9.5.


### What did you see happen?

v2.9.6 generates (incorrect):
  SELECT a.id, a.name
  FROM "demo_a" AS a
  LEFT JOIN "demo_b" AS "b" ON (a.id=b.data_id)
  ORDER BY "demo_a".sort
  LIMIT 10 OFFSET 0

  v2.9.6 incorrectly adds the original table name prefix to the column in ORDER BY, instead of:
  1. Keeping just the column name (as in v2.9.5), OR
  2. Using the alias a.sort if a prefix is needed

  Error Message (PostgreSQL)

  pq: invalid reference to FROM-clause entry for table "demo_a"
  HINT: Perhaps you meant to reference the table alias "a".

  Root Cause Analysis

  This issue was likely introduced by PR #4521 "Resolve column ambiguity in GROUP BY/ORDER BY with MySQL JOIN".

  The fix correctly tries to add table prefixes to avoid ambiguity, but it uses the original table name instead of the alias set by .As().


  v2.9.5 generates (correct):
  SELECT a.id, a.name
  FROM "demo_a" AS a
  LEFT JOIN "demo_b" AS "b" ON (a.id=b.data_id)
  ORDER BY "sort"
  LIMIT 10 OFFSET 0

  The Order() should generate ORDER BY "sort" (column name only), as it did in v2.9.5.

 - If .As("a") is called, any auto-prefixing should use a.column, not original_table.column
  - Or simply keep the column name as-is when it's unambiguous (v2.9.5 behavior)

Suggested Fix

  When auto-prefixing columns in Order()/Group(), the framework should:
  1. Check if an alias was set via .As()
  2. If yes, use the alias as prefix (e.g., a.sort)
  3. If no alias, use the original table name

  Workaround

  Manually specify the alias prefix:

  // Instead of:
  Order(dao.DemoA.Columns().Sort)

  // Use:
  Order("a.sort")

lingcoder:
## What this PR does

  Revert the auto table prefix behavior in `Order()` and `Group()` introduced by #4521.

  ## Why

  PR #4521 attempted to resolve column ambiguity in GROUP BY/ORDER BY with MySQL JOIN by automatically adding table prefixes to unqualified columns. However,
  this approach has issues:

  1. When using `.As()` to set table alias, it uses the original table name instead of the alias, causing errors in PostgreSQL and other databases
  2. The framework cannot reliably determine which table the user intends when multiple tables have the same column
  3. Adds hidden behavior that users may not expect

  ## Example of the issue (#4554)

  ```go
  db.Model("demo_a").As("a").
      LeftJoin("demo_b", "b", "a.id=b.data_id").
      Order("sort").All()

  Expected (v2.9.5):
  ORDER BY "sort"

  Actual (v2.9.6):
  ORDER BY "demo_a".sort  -- Wrong! Should use alias "a" or no prefix

  Solution

  Revert to v2.9.5 behavior: Order("sort") generates ORDER BY "sort" without auto-prefixing. Users should explicitly specify table prefix when needed:
  Order("a.sort").

  Closes #4554
  ```


gqcn:
Thanks for your suggestion! We'll carefully consider and test it, though this may take a little time. We also welcome other guys who see this to join in code review and participate in the discussion.

hailaz:
撤销 #4521 的更改，自动前缀提高维护的复杂度，减少了透明度，增加了调试的难度。手动设置前缀更加直观和可预测。

PR #4557, opened 2025-12-10
Update pgsql_format_upsert.go

ritterrong-dot:
**What changes were proposed?**
Modified the `FormatUpsert` method in the PostgreSQL driver (`driver_pgsql.go`). Previously, when the `option.OnConflict` slice was empty, the method could return an error. Now, it delegates the call to the parent class's `FormatUpsert` method (`d.Core.FormatUpsert`) instead.

**Why are these changes needed?**
This fix ensures consistency and extensibility. The original behavior prevented the generic upsert logic (which may be implemented differently by other database drivers like My custom pgSQL with `ON DUPLICATE KEY UPDATE`) from being executed. The parent class should have the opportunity to handle cases where PostgreSQL-specific conflict columns are not specified.

### DIFF
--- a/contrib/drivers/pgsql/pgsql_format_upsert.go
+++ b/contrib/drivers/pgsql/pgsql_format_upsert.go
@@ -10,8 +10,8 @@ import (
 	"fmt"
 
 	"github.com/gogf/gf/v2/database/gdb"
-	"github.com/gogf/gf/v2/errors/gcode"
-	"github.com/gogf/gf/v2/errors/gerror"
+	// "github.com/gogf/gf/v2/errors/gcode"
+	// "github.com/gogf/gf/v2/errors/gerror"
 	"github.com/gogf/gf/v2/text/gstr"
 	"github.com/gogf/gf/v2/util/gconv"
 )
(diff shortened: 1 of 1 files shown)


ritterrong-dot:
// Or use the solution

// It automatically infers and uses the table's primary key as the conflict column
// when `option.OnConflict` is not explicitly provided.
//
// Design Rationale:
// This enhances API usability in the common scenario where the target table has a
// primary key and the data to be saved includes the primary key's value.
// Previously, the API forced users to manually specify `.OnConflict("primary_key")`
// even when it could be unambiguously determined, resulting in unnecessary boilerplate.
// This method now allows users to simply call `Model.Data(data).Save()` in such cases.
func (d *Driver) handleUpsert(ctx context.Context, table string, data interface{}, option *gdb.DoInsertOption) error {
    // 1. Proceed only if the user has not manually specified conflict columns.
    if len(option.OnConflict) == 0 {
        // 2. First, check if the primary key information is already available in the context (e.g., from a previous cache or operation).
        if pkField, ok := ctx.Value(internalPrimaryKeyInCtx).(gdb.TableField); ok && pkField.Name != "" {
            option.OnConflict = []string{pkField.Name}
        } else {
            // 3. Context cache miss. Query the database for the table's schema to locate the primary key.
            tableFields, err := d.GetCore().GetDB().TableFields(ctx, table)
            if err != nil {
                // The Upsert operation's target is undefined without a primary key. Fail with a descriptive error.
                return gerror.Wrapf(err, "failed to query table fields to resolve primary key for upsert")
            }
            // 4. Scan the table fields to find the column marked as primary key ('pri').
            for _, field := range tableFields {
                if field.Key == "pri" {
                    option.OnConflict = []string{field.Name}
                    // Optional: Cache this primary key field in the context for the remainder of this operation.
                    // ctx = context.WithValue(ctx, internalPrimaryKeyInCtx, *field)
                    break
                }
            }
            // 5. If the loop completes without finding a primary key, the table lacks a viable unique constraint for an Upsert.
            if len(option.OnConflict) == 0 {
                return gerror.NewCode(gcode.CodeInvalidRequest, "table %s has no primary key defined, which is required for the upsert operation", table)
            }
        }
    }
    // ... Normal Upsert processing continues here, using the resolved `option.OnConflict`.
    return nil
}

ritterrong-dot:
Solve it in another way

PR #4600, opened 2026-01-09
fix(util/gconv): gconv unsafe str to bytes

liov:
The gconv.UnsafeStrToBytes function has been updated to use the Go 1.20+ safe approach, as the previous implementation could cause a panic in certain scenarios.

For example, when an HTTP request header specifies Content-Type: application/x-www-form-urlencoded, but the actual request body contains JSON data, the following code attempts to detect and handle this case:
```go
if !gregex.IsMatchString(`^[\w\-\[\]]+$`, name) && len(r.PostForm) == 1 {
    // It might be JSON/XML content.
    if s := gstr.Trim(name + strings.Join(values, " ")); len(s) > 0 {
        if s[0] == '{' && s[len(s)-1] == '}' || s[0] == '<' && s[len(s)-1] == '>' {
            r.bodyContent = gconv.UnsafeStrToBytes(s)
            params = ""
            break
        }
    }
}
```
However, after this assignment, bodyContent ends up with a capacity (cap) of 0. slice operations like [:] perform stricter validation and will panic if the capacity is 0. This causes a panic in functions such as:

```go
body = bytes.TrimSpace(body)

func TrimSpace(s []byte) []byte {
    ...
    return s[start:stop] // panic here due to cap == 0
}
```
The capacity (cap) of the slice returned by directly calling this function is unpredictable, as it depends on the adjacent memory layout. However, within the framework, this causes issues—likely because, starting from Go 1.22, the standard library's parseForm implementation consistently appends a trailing zero byte after the string data in memory.
This PR fix the problem.

------------------------------------
gconv unsafe str to bytes 改用 go1.20 后的写法，之前的写法在某些场景下会 panic
例如 http 请求头为`application/x-www-form-urlencoded`,实际的 body 为 json,
经过解析后
```go
	if !gregex.IsMatchString(`^[\w\-\[\]]+$`, name) && len(r.PostForm) == 1 {
					// It might be JSON/XML content.
					if s := gstr.Trim(name + strings.Join(values, " ")); len(s) > 0 {
						if s[0] == '{' && s[len(s)-1] == '}' || s[0] == '<' && s[len(s)-1] == '>' {
							r.bodyContent = gconv.UnsafeStrToBytes(s)
							params = ""
							break
						}
					}
				}
```
bodyContent的 cap 为 0，由于切片操作[:]会校验 cap 为 0，会直接 panic
```go
body = bytes.TrimSpace(body)

---
func TrimSpace(s []byte) []byte {
...
return s[start:stop] // panic
}
```
直接使用这个函数得到的 cap 会是随机的, 因为跟的内存不确定，但是在框架中有问题，估计是1.22 后标准库parseForm 的时候后面内存固定跟了个 0
该 PR 修复这个问题

### DIFF
--- a/util/gconv/gconv_unsafe.go
+++ b/util/gconv/gconv_unsafe.go
@@ -12,12 +12,12 @@ import "unsafe"
 // Note that, if you completely sure you will never use `s` variable in the feature,
 // you can use this unsafe function to implement type conversion in high performance.
 func UnsafeStrToBytes(s string) []byte {
-	return *(*[]byte)(unsafe.Pointer(&s))
+	return unsafe.Slice(unsafe.StringData(s), len(s))
 }
 
 // UnsafeBytesToStr converts []byte to string without memory copy.
 // Note that, if you completely sure you will never use `b` variable in the feature,
 // you can use this unsafe function to implement type conversion in high performance.
 func UnsafeBytesToStr(b []byte) string {
-	return *(*string)(unsafe.Pointer(&b))
+	return unsafe.String(unsafe.SliceData(b), len(b))
 }


gqcn (approved):
/lgtm

PR #4701, opened 2026-02-13
fix(databse/gdb): use COUNT(1) if fields number is greater than 1 even when parameter `useFieldForCount` is true in AllAndCount/ScanAndCount

### LINKED ISSUE #4698: database/gdb: AllAndCount with multiple fields generates invalid COUNT syntax
## Go version
go version go1.25.0 linux/amd64

## GoFrame version
v2 (master branch, commit sha: latest)

## Can this bug be reproduced with the latest release?
Option Yes

## What did you do?
Called `AllAndCount(true)` with multiple fields specified:

```go
result, count, err := db.Model(table).Fields("id, nickname").AllAndCount(true)
```

**Test case reference:** `Test_Model_AllAndCount_MultipleFields_WithTotal` in PR #4703

## What did you see happen?
The framework generated invalid COUNT SQL with field list instead of `COUNT(*)`:

```sql
SELECT COUNT(id, nickname) FROM `user_xxx`
             ^^ Invalid - COUNT takes 1 argument
```

SQL execution error:
```
Error 1064 (42000): You have an error in your SQL syntax ... near ', nickname) FROM ...'
```

**Root cause:** When `AllAndCount(true)` is used with multiple fields, the COUNT query incorrectly includes the field list.

## What did you expect to see?
When using `AllAndCount(true)` with any fields configuration:

```sql
-- Data query (correct)
SELECT id, nickname FROM `user_xxx`

-- Count query (should always use COUNT(*) or COUNT(1))
SELECT COUNT(*) FROM `user_xxx`
```

The count query should **always** use `COUNT(*)` or `COUNT(1)`, regardless of the fields specified in the data query.

## Additional Context
- **Parent issue:** #4689 (comprehensive database driver test coverage)
- **Discovered in:** PR #4703 while adding pagination tests
- **Affected drivers:** All database drivers
- **Regression test:** Added in PR #4703

lingcoder:
## Summary
Fix bug where `AllAndCount(true)` with multiple fields generates invalid SQL `COUNT(field1, field2, ...)` causing syntax error.

## Root Cause
When `useFieldForCount=true`, the COUNT query inherits the fields configuration from the model:
```go
// Before (buggy code)
if !useFieldForCount {
    countModel.fields = []any{Raw("1")}
}
// When useFieldForCount=true, fields remain as ["id", "nickname"]
// Generates: SELECT COUNT(id, nickname) FROM table ❌
```

## Fix
Always use `COUNT(1)` regardless of `useFieldForCount` parameter since COUNT() accepts only one argument:
```go
// After (fixed code)
// Always use COUNT(1) for counting, regardless of useFieldForCount.
// COUNT() accepts only one argument, so we can't use multiple fields.
countModel.fields = []any{Raw("1")}
```

Applied to both `AllAndCount()` and `ScanAndCount()` methods.

## Tests
Added `Test_Issue4698` with 5 test cases:
1. AllAndCount(true) with multiple fields
2. AllAndCount(false) with multiple fields (baseline)
3. ScanAndCount with multiple fields
4. AllAndCount with single field
5. AllAndCount with WHERE condition

All tests verify that COUNT generates valid SQL and returns correct count.

## Related
Fixes #4698
Ref #4703 (discovered during pagination test development)

### DIFF
--- a/contrib/drivers/mysql/mysql_z_unit_issue_test.go
+++ b/contrib/drivers/mysql/mysql_z_unit_issue_test.go
@@ -1983,3 +1983,86 @@ func Test_Issue4697(t *testing.T) {
 		t.AssertNE(result[0]["nickname"], nil)
 	})
 }
+
+// https://github.com/gogf/gf/issues/4698
+func Test_Issue4698(t *testing.T) {
+	table := createInitTable()
+	defer dropTable(table)
+
+	// Test 1: AllAndCount with multiple fields should generate valid COUNT SQL
+	gtest.C(t, func(t *gtest.T) {
+		result, count, err := db.Model(table).Fields("id, nickname").AllAndCount(true)
+		t.AssertNil(err)
+		t.Assert(count, TableSize)
+		t.Assert(len(result), TableSize)
+		t.AssertNE(result[0]["id"], nil)
+		t.AssertNE(result[0]["nickname"], nil)
+		t.Assert(result[0]["passport"], nil)
+	})
+
+	// Test 2: AllAndCount(false) with multiple fields
+	gtest.C(t, func(t *gtest.T) {
+		result, count, err := db.Model(table).Fields("id, nickname").AllAndCount(false)
+		t.AssertNil(err)
+		t.Assert(count, TableSize)
+		t.Assert(len(result), TableSize)
+	})
+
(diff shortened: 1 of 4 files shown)


lingcoder:
已按要求整改

gqcn (approved):

Issue #4689, opened 2026-02-11
contrib/drivers: comprehensive database driver test coverage improvement plan


lingcoder:
### Current State

GoFrame's `contrib/drivers` provides 11 database driver implementations.
Each driver adapts `gdb.Model`'s 161 public methods to its specific SQL dialect.
MySQL driver has the most mature test suite, but there's a significant gap between drivers.

> Data as of: 2026-02-12

| Driver | Test Files | Test Functions | Test Lines | Assessment |
|--------|-----------|---------------|-----------|------------|
| MySQL | 23 | 348 | ~18,700 | Broad coverage, depth gaps in some areas |
| MariaDB | 2 | 39 | ~900 | Minimal, MySQL-compatible |
| PgSQL | 11 | 110 | ~5,500 | +5 PRs adding ~14 files |
| GaussDB | 13 | 123 | ~5,800 | +4 PRs adding ~13 files |
| SQLite | 3 | 157 | ~4,200 | High function count but few files |
| MSSQL | 5 | 139 | ~4,000 | Missing transaction/where/hook coverage |
| Oracle | 3 | 60 | ~1,800 | Significant gaps |
| DM | 6 | 39 | ~1,500 | Significant gaps |
| ClickHouse | 5 | 31 | ~1,200 | Limited by column-store nature |
| OceanBase | 0 | 0 | 0 | No tests |
| TiDB | 0 | 0 | 0 | No tests |

### Problem

1. **Database driver coverage imbalance**: MySQL has 23 test files, while most database drivers have 2~6, making it hard to guarantee driver adapter quality across databases
2. **No shared baseline**: Each database driver's tests were written independently, no systematic standard to measure coverage against
3. **MySQL itself has depth gaps**: Despite being the most complete, Lock operations (4 variants), OmitEmpty/OmitNil (6 methods), Context timeout/cancellation, concurrency safety, and advanced data types (JSON/Binary/Decimal) lack dedicated testing in the MySQL driver
4. **Duplicated effort**: Similar database operation tests are rewritten per driver rather than systematically ported

## Core Idea

**MySQL First → Standardized Baseline → Per-Driver Alignment + Driver-Specific Features**

Use MySQL database driver as the single source of truth. First make MySQL driver tests comprehensive, then systematically port them to all other database drivers using standard SQL, and finally add driver-specific feature tests for each database's unique capabilities.

### Three-Layer Test Model

| Layer | Scope | Content |
|-------|-------|---------|
[...]

PR #4732, opened 2026-03-03
test(contrib/drivers/mysql): add nil check in Test_DB_Ctx

lingcoder:
## Summary

Add `t.AssertNE(err, nil)` before `err.Error()` in `Test_DB_Ctx` to prevent potential nil pointer panic if the context-timeout query unexpectedly succeeds.

## Changes

- `contrib/drivers/mysql/mysql_z_unit_core_test.go`: Add nil check for err before accessing err.Error() in Test_DB_Ctx

ref #4689


lingcoder:
放在 PR 4719里一起提交了。改动很小，主要为了让copilot闭嘴= =！

PR #4745, opened 2026-03-31
feat(database/gdb): support multiple chained hooks

UncleChair:
1. Remove Next() function to prevent users controlling hook process, in case the hook chain is blocked by some wrong usage with missing in.Next(), the new 'next' process is done internally
2. Support multiple hook register in Hook() function and added chain calling support for multiple hooks
3. Updated related test cases
4. fixed some lint problems


gqcn:
I cannot complete this review reliably yet, so it would help to have a maintainer take a look.

The reason is: `AGENTS.md` is not present on the PR's target commit, so the code cannot be checked against the repository rules from that base.

Suggested focus: multiple chained hooks on gdb.Model; currently a draft.

Suggested reviewers, if available: @wln32 @fainc @oldme-git

Why they are mentioned: they have worked on related files.

PR #4756, opened 2026-04-16
feat(contrib/drivers/mssql): convert uniqueidentifier to uuid.UUID

NickWilde18:
## Summary

SQL Server `UNIQUEIDENTIFIER` columns arrive on the wire as 16 bytes whose first 8 follow the little-endian COM/Win32 GUID layout while the remaining 8 are big-endian. The MSSQL driver previously had no override for this column type, so `gdb`'s default conversion path treated the raw bytes as a string and returned garbage characters. Any caller doing `row[\"id\"].String()` on a `uniqueidentifier` column would get mojibake.

This PR mirrors the existing pgsql `uuid` handling: add `CheckLocalTypeForField` and `ConvertValueForLocal` overrides on the MSSQL driver that detect `uniqueidentifier` (case-insensitive), delegate the byte-order swap to `mssql.UniqueIdentifier.Scan`, and return a `uuid.UUID`. All other column types fall through to `Core` unchanged.

## Reproduction (before this PR)

\`\`\`sql
CREATE TABLE t_uuid (id UNIQUEIDENTIFIER NOT NULL DEFAULT NEWID(), label NVARCHAR(50));
INSERT INTO t_uuid (label) VALUES ('hello');
\`\`\`

\`\`\`go
rows, _ := db.Model(\"t_uuid\").Ctx(ctx).Fields(\"id, label\").All()
for _, r := range rows {
    fmt.Println(r[\"id\"].String()) // garbage characters like \"\ufffdԓ\ufffd?\\\"\ufffdB\ufffdGx\ufffd...\"
}
\`\`\`

After this PR the same call returns the canonical UUID string (e.g. `da93d4f6-223f-42b2-a647-789371ffa693`).

## Tests

- `Test_CheckLocalTypeForField_UniqueIdentifier` — verifies `uniqueidentifier` (and `UNIQUEIDENTIFIER`) maps to `LocalTypeUUID`, and that unknown types still fall through to Core (`nvarchar(100) → LocalTypeString`).
- `Test_ConvertValueForLocal_UniqueIdentifier` — covers the wire-format byte swap, the string form (e.g. after a server-side `CAST AS NVARCHAR(36)`), a sanity check against `mssql.UniqueIdentifier.Scan` directly, and the invalid-length error path.

These unit tests do not require a running MSSQL instance — they exercise the driver hooks against fixture inputs. Existing DB-backed tests continue to use the CI's MSSQL container unchanged.

## Notes

- `github.com/google/uuid` is promoted from indirect to direct dependency in `contrib/drivers/mssql/go.mod` (already pulled in transitively via `microsoft/go-mssqldb`).
- This patch concerns only the read path (DB → Go). Write/where-clause paths (Go → DB) already work because MSSQL accepts the canonical string form server-side.


NickWilde18:
Thanks @copilot-pull-request-reviewer for the review. Both points addressed in 8ecf492:

1. Renamed the `mssql` import alias to `mssqldriver` in `mssql_convert.go` so it no longer collides with the package name and matches what the test file already does.
2. Replaced the misleading fallback in the `Scan` error branch with `return nil, gerror.Wrapf(err, ...)`, matching the pgsql driver's pattern. The fallback string was unreachable as you noted.

Removed the now-unused `fmt` and `gconv` imports.

hailaz:
@copilot resolve the merge conflicts in this pull request

hailaz (approved):

Issue #4697, opened 2026-02-13
database/gdb: Fields with empty string generates invalid SQL


lingcoder:
## Go version
go version go1.25.0 linux/amd64

## GoFrame version
v2 (master branch, commit sha: latest)

## Can this bug be reproduced with the latest release?
Option Yes

## What did you do?
Called `Fields("")` with an empty string on a database Model:

```go
result, err := db.Model(table).Fields("").Limit(1).All()
```

**Test case reference:** `Test_Model_Fields_Empty` in PR #4703

## What did you see happen?
The framework generated invalid SQL with missing field list:

```sql
SELECT  FROM `user_1770956455090062620` LIMIT 1
       ^^ Missing fields here
```

SQL execution error:
```
Error 1064 (42000): You have an error in your SQL syntax ... near 'FROM `user_1770956455090062620` LIMIT 1'
```

## What did you expect to see?
Empty string in Fields should either:
- **Option A (Recommended):** Be ignored (equivalent to not calling Fields), selecting all fields: `SELECT * FROM ...`
- **Option B:** Return a clear parameter validation error

## Additional Context
- **Parent issue:** #4689 (comprehensive database driver test coverage)
- **Discovered in:** PR #4703 while adding pagination and error handling tests
- **Affected drivers:** All database drivers (MySQL, PostgreSQL, etc.)
- **Regression test:** Added in PR #4703 to prevent recurrence after fix

Issue #4677, opened 2026-02-06
record.Get().Bytes() corrupts bytea data on retrieval from PostgreSQL


RidaLiu:
### Go version

go1.25.0 darwin/arm64

### GoFrame version

v2.9.4

### Can this bug be reproduced with the latest release?

Option Yes

### What did you do?

 When storing a binary []byte slice (such as a gob stream) in a PostgreSQL bytea column and
   retrieving it using the gogf/gf ORM, the record.Get("column").Bytes() method does not    
  return a byte-for-byte identical slice. It appears to append extra, unexpected data to the
   end of the byte slice, causing deserialization errors for binary formats like gob.
  Environment:                                                                              
  - Database: PostgreSQL                                                                    
  - Column Type: bytea 

### What did you see happen?

Steps to Reproduce:                                                                       
  1. Save a []byte slice to a bytea column using dao.Data(do.MyStruct{Value:                
  myBytes}).Save().                                                                         
  2. Retrieve the row using dao.Where(...).One().                                           
  3. Get the value using retrievedBytes := record.Get("value").Bytes().                     
  4. Compare the len() and sha256.Sum256() of myBytes and retrievedBytes.                   
                                                                            

### What did you expect to see?

Expected Result:                                                                          
  The length and SHA256 hash of the original and retrieved byte slices should be identical. 
 Actual Result:                                                                            
  The retrieved byte slice is longer than the original and has a different SHA256 hash,     
  indicating data corruption.

lingcoder:
分析修复了下

RidaLiu:
> 分析修复了下

一般多久 pr 会被 approve？

lingcoder:
> > 分析修复了下
> 
> 一般多久 pr 会被 approve？

@gqcn @hailaz

Issue #4595, opened 2026-01-08
FieldsPrefix silently drops fields when using table alias before LeftJoin (PostgreSQL)


lingcoder:
### Go version

go1.25.5

### GoFrame version

v2.9.6

### Can this issue be reproduced with the latest release?

Yes

### What did you do?

When using multiple `FieldsPrefix` calls with different table aliases before `LeftJoin`, fields from the first aliased table are silently dropped.

**DDL for reproduction (PostgreSQL):**

```sql
CREATE TABLE "user" (
    id BIGSERIAL PRIMARY KEY,
    name VARCHAR(100),
    email VARCHAR(100)
);

CREATE TABLE "user_detail" (
    id BIGSERIAL PRIMARY KEY,
    user_id BIGINT,
    phone VARCHAR(20),
    address VARCHAR(200)
);
```

**Code to reproduce:**

```go
result, err := g.DB().Model("user").As("t1").
    FieldsPrefix("t2", "phone", "address").
    FieldsPrefix("t1", "id", "name", "email").
    LeftJoin("user_detail", "t2", "t1.id=t2.user_id").
    All(ctx)
```

### What did you expect to see?

```sql
SELECT "t2"."phone", "t2"."address", "t1"."id", "t1"."name", "t1"."email" 
FROM "user" AS t1 
LEFT JOIN "user_detail" AS "t2" ON (t1.id=t2.user_id)
```

### What did you see instead?

```sql
SELECT "t1"."id", "t1"."name", "t1"."email" 
FROM "user" AS t1 
LEFT JOIN "user_detail" AS "t2" ON (t1.id=t2.user_id)
```

The `t2` fields (`phone`, `address`) are completely missing from the SELECT clause.

### Workaround

Use `Fields()` with manual prefix:

```go
g.DB().Model("user").As("t1").
    Fields("t2.phone", "t2.address", "t1.id", "t1.name", "t1.email").
    LeftJoin("user_detail", "t2", "t1.id=t2.user_id")
```

PR #4763, opened 2026-04-23
test(contrib/drivers/pgsql): expand test coverage to MySQL/MariaDB baseline + 3 framework fixes

lingcoder:
## Summary

Comprehensive PgSQL driver test coverage expansion, porting ~500 test functions from the MySQL/MariaDB baseline, adding PgSQL-specific Layer 3 tests (JSONB/RETURNING/CTE/window functions/ARRAY), and fixing 3 framework-level bugs discovered during porting.

**Scope**: 21 commits / 47 files / ~+10750 / -460
**Test results**: 560 PASS / 0 FAIL / 29 SKIP (all skips are PgSQL-incompatible MySQL features like REPLACE INTO / ORDER BY NULL / HAVING without GROUP BY)

## Framework Bugs Discovered & Fixed

Three bugs surfaced during test porting. All follow the canonical `OrderRandomFunction()` driver-dispatch pattern — new interface methods on `DB`, defaults on `Core` preserve MySQL legacy behavior, specific drivers override their dialect.

### 1. Boolean soft-delete `col=0` on strict-bool drivers
- **Where**: `database/gdb/gdb_model_soft_time.go` — `buildDeleteCondition` hardcoded `col=0` for `LocalTypeBool`, which fails on PgSQL/GaussDB/ClickHouse's strict boolean type.
- **Fix**: New `DB.GetBoolLiteral(v bool) string`. Core default returns `"1"`/`"0"`; pgsql/gaussdb/clickhouse override to `"true"`/`"false"`. Soft-delete condition builder now calls `m.db.GetBoolLiteral(false)` instead of branching on driver name.

### 2. `LockShared()` hardcodes MySQL `LOCK IN SHARE MODE`
- **Where**: `database/gdb/gdb_model_lock.go` — `LockShared()` emitted MySQL-only syntax regardless of target driver, causing PgSQL/GaussDB errors.
- **Fix**: New `DB.GetLockSharedClause() string`. Core default returns `"LOCK IN SHARE MODE"`; pgsql/gaussdb override to `"FOR SHARE"`. `Model.LockShared()` now calls `m.db.GetLockSharedClause()`. User code using `LockShared()` works transparently on all drivers.

### 3. `CatchSQL`/`ToSQL` silently drop INSERT...RETURNING SQL
- **Where**: `contrib/drivers/pgsql/pgsql_do_exec.go` + `contrib/drivers/gaussdb/gaussdb_do_exec.go` — manually called `FormatSqlBeforeExecuting` + `DoFilter` + `DoCommit` on the `InsertAndGetId` path, bypassing `Core.DoQuery`'s `CatchSQLManager` handling. This broke two APIs:
  - `CatchSQL` silently dropped the RETURNING SQL from captured array
  - `ToSQL` ignored `DoCommit=false` and actually executed the INSERT
- **Fix**: Delegate to `Core.DoQuery` so both behaviors are honored uniformly. Regression tests `Test_PgSQL_Returning_CatchSQL` and `Test_PgSQL_Returning_ToSQL` added.

## Test Coverage Breakdown

| Layer | Files | Notes |
|-------|-------|-------|
| **Core** | `core_test.go` | ClearTableFields etc. |
| **Model API** | `model_extra_test.go` | Full chained API coverage ported from MariaDB |
| **Transaction** | `transaction_test.go` | Propagation/Isolation/Deadlock/Nested (71 funcs) |
| **Features** (ported) | 14 files | batch/cache/concurrent/ctx/duplicate/error/hook/lock/metadata/join/subquery/omit/pagination/partition/raw-type/scanlist/soft-time/with |
| **Issues** | `issue_test.go` | 47 historical regression tests |
| **Layer 3** (PgSQL-specific) | `feature_pgsql_test.go` | JSONB operators/RETURNING/CTE/window/ARRAY/DISTINCT ON |

## Historical Issues Regression-Tested

The PR adds regression tests covering 46 previously-fixed issues to prevent regression on the PgSQL driver:

`#1002` `#1380` `#1401` `#1412` `#1570` `#1700` `#1701` `#1733` `#1934` `#2012` `#2105` `#2119` `#2231` `#2338` `#2339` `#2356` `#2427` `#2439` `#2552` `#2561` `#2643` `#2782` `#2787` `#2907` `#3086` `#3204` `#3218` `#3238` `#3330` `#3626` `#3632` `#3649` `#3668` `#3671` `#3754` `#3915` `#3932` `#3968` `#4033` `#4034` `#4086` `#4231` `#4500` `#4595` `#4677` `#4697` `#4698`

## Non-bug adaptations

- `Test_Issue1412`: MySQL original uses `with:Id=ItemId` (Go field name on LHS) which only works due to MySQL's case-insensitive identifiers. Rewrote the PgSQL port to use the DB column name (`with:id=ItemId`) per GoFrame's "what you write is what you get" design philosophy. All other With tests already use canonical DB column names — this is not a framework bug.

## Test plan

- [x] All 560 tests pass against local PgSQL 15 in Docker
- [x] gofmt / gci / go vet all clean
- [x] All contrib drivers (mysql/mariadb/pgsql/gaussdb/clickhouse/oracle/mssql/dm/sqlite/oceanbase/tidb) build cleanly — Core defaults preserve existing behavior for non-overriding drivers
- [x] GaussDB soft-time tests synchronized with the framework fix (no follow-up PR needed for gaussdb)
- [x] Rebased onto latest master

ref #4689


hailaz (approved):

lingcoder:
@gqcn  看下

PR #4777, opened 2026-05-27
feat(os/gapp): add unified multi-server application lifecycle

qinyuguang:
Introduce os/gapp App to coordinate Boot, Start, Stop, and OS signal handling across multiple Server implementations, supporting graceful or forceful shutdown in one place.

- Add contrib/scheduler/gjob worker and cron servers implementing gapp.Server
- Add grpcx GrpcServerAdapter with managed non-blocking start
- Add ghttp.Server.Close for forceful shutdown; harden admin restart/shutdown
- Add gproc.StartListen for background signal handling used by gapp

## Example

```go
// main initializes and starts gRPC/HTTP servers and scheduler job servers (worker + cron).
func main() {
	ctx := gctx.GetInitCtx()

	grpcServer := grpcx.Server.New()
	controller.Register(grpcServer)

	httpServer := g.Server()
	httpServer.BindHandler("/", func(r *ghttp.Request) {
		r.Response.Write("Hello World!")
	})

	app := gapp.New(
		grpcx.NewGappServerAdapter(grpcServer),
		gapp.NewHTTPServerAdapter(httpServer),
	)
	app.Add(gjob.NewServersFromConfig(ctx, job.Handlers())...)

	app.Option(
		gapp.NewOption(func(ctx context.Context, app *gapp.App) {
			g.Log().Info(ctx, "app starting...")
		}),
	)

	if err := app.Run(ctx); err != nil {
		g.Log().Errorf(ctx, "app run failed: %+v", err)
		os.Exit(1)
	}

	g.Log().Info(ctx, "app shutdown.")
}

```


gqcn:
`os/gapp` plus the gjob/grpc adapters is a sizable new lifecycle API, and the bilingual READMEs plus unit tests are a strong start. Two items would help before merge:

- **Suggested priority** `contrib/rpc/grpcx/grpcx_grpc_server.go` (`doServeAsynchronously`): a failed `Server.Serve` used to `Fatalf` and stop the process. It now only logs, so standalone `Run()` can sit in `doSignalListen` with a dead listener. Consider keeping `Fatalf` on the blocking `Run()` path, and using `Errorf` only for `StartManaged`.
- **Suggested improvement** `contrib/rpc/grpcx/grpcx_gapp.go`: this is not the package's main file, but it starts with a `Package grpcx ...` comment. Consider a file-purpose comment for the adapter instead.

Issue #4792, opened 2026-07-07
[gsession] StorageFile.SetSession 非原子写入导致并发读请求看到空 session


windf17:
## Description

`StorageFile.SetSession` writes session data using a non-atomic file write pattern (open with `O_TRUNC` then two separate `Write` calls), creating a race window where concurrent readers see an empty or partially-written file.

## Impact

Under concurrent requests sharing the same session ID (common in SPA applications where page switching fires multiple API calls simultaneously), intermittent authentication failures occur:

- Session data is lost; `GetSession` returns nil
- User appears "not logged in" and gets redirected to login

### Write path (`gsession_storage_file.go:189-203`)

```go
file, err := gfile.OpenWithFlagPerm(
    path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, os.ModePerm,
)
// ...
file.Write(gbinary.EncodeInt64(gtime.TimestampMilli()))  // 8 bytes timestamp
file.Write(content)                                         // JSON body
```

`O_TRUNC` empties the file immediately. Then **two separate `file.Write()` calls** create a real gap between writing the timestamp header and the JSON body.

### Read path (`gsession_storage_file.go:141-170`)

```go
content = gfile.GetBytes(path)
if len(content) > 8 {
    // parse JSON
}
return nil, nil  // content <= 8 bytes → empty session
```

### Race window

```
Writer(SetSession)                  Reader(GetSession)
│                                    │
├─O_TRUNC open → file is empty      │
│                                    ├─gfile.GetBytes() → 0-8 bytes
│                                    │ len<=8 → returns nil session
│                                    │ → user appears not logged in!
├─file.Write(timestamp)[8 bytes]    │
├─file.Write(content)[JSON body]    │
│ write complete                     │
```

## Production Evidence

Real log (3ms window, same session ID):

```
12:16:47.467  devices/all          userId=1  ← normal
12:16:47.470  software-systems/all userId=0  ← user_id key missing! (race hit)
12:16:47.472  network-lines/all    userId=1  ← normal
```

## Proposed Fix: atomic write via temp file + rename

```go
func (s *StorageFile) SetSession(ctx context.Context, sessionId string, sessionData *gmap.StrAnyMap, ttl time.Duration) error {
    intlog.Printf(ctx, "StorageFile.SetSession: %s, %v, %v", sessionId, sessionData, ttl)
    path := s.sessionFilePath(sessionId)
    content, err := json.Marshal(sessionData)
    if err != nil {
[...]

PR #4799, opened 2026-07-20
fix(os/gsession): serialize session file IO under per-id lock

sonnemusk:
Fixes concurrent file-session corruption when `GetSession` races `SetSession` on the same id (#4792).

`StorageFile.SetSession` opens with `O_TRUNC` then writes the 8-byte timestamp and JSON body. Without synchronization a concurrent `GetSession` can read a truncated file (`len <= 8`) and treat the session as missing.

Fix: per-`sessionId` `sync.RWMutex` around Get/Set/Remove file IO. Write path stays the same shape as master (no temp+rename / no large buffer alloc — those tripped CodeQL).

```
go test ./os/gsession/ -run StorageFile -count=1
```

### DIFF
--- a/os/gsession/gsession_storage_file.go
+++ b/os/gsession/gsession_storage_file.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


sonnemusk:

PR #4800, opened 2026-07-20
fix(database/gdb): Result.Structs fill []map destinations

### LINKED ISSUE #4787: gdb: `Result.Structs` / `Model.Scan` into `*[]map[string]interface{}` silently returns empty maps instead of data or error
## Version

- `github.com/gogf/gf/v2 v2.9.0`
- `github.com/gogf/gf/contrib/drivers/mysql/v2 v2.8.3`
- go 1.23, linux/amd64
- (Confirmed the relevant code path is unchanged on `master` as of this writing.)

## What happens

`db.Model(...).Scan(&rows)` / `db.Raw(sql).Scan(&rows)` where `rows` is `[]map[string]interface{}`
returns a slice of the **correct length** but every element is an **empty map**, and **no error**
is returned. Switching the destination to a typed struct slice, or reading via `.All()` /
`.All().List()`, works correctly against the very same query.

## Minimal, self-contained reproduction (no database needed)

The bug is in the result→slice conversion, so it reproduces against a hand-built `gdb.Result`:

```go
package main

import (
	"fmt"

	"github.com/gogf/gf/v2/container/gvar"
	"github.com/gogf/gf/v2/database/gdb"
)

func main() {
	// A gdb.Result is []gdb.Record, i.e. []map[string]*gvar.Var — exactly what a driver returns.
	res := gdb.Result{
		gdb.Record{"regionId": gvar.New(1)},
		gdb.Record{"regionId": gvar.New(2)},
		gdb.Record{"regionId": gvar.New(3)},
	}

	// (1) Result.Structs into *[]map  — what Model.Scan(&[]map) dispatches to.
	var viaStructs []map[string]interface{}
	_ = res.Structs(&viaStructs)
	fmt.Printf("Result.Structs : len=%d first=%v\n", len(viaStructs), first(viaStructs))
	// => len=3 first=map[]   <-- BUG: 3 empty maps, no error

	// (2) gconv.Scan on the same source into *[]map — works.
	var viaScan []map[string]interface{}
	_ = gconv.Scan(res, &viaScan)
	fmt.Printf("gconv.Scan     : len=%d first=%v\n", len(viaScan), first(viaScan))
	// => len=3 first=map[regionId:1]

	// (3) Result.List() — works.
	fmt.Printf("Result.List()  : len=%d first=%v\n", len(res.List()), first(res.List()))
	// => len=3 first=map[regionId:1]
}

func first(s []map[string]interface{}) interface{} {
	if len(s) == 0 {
		return nil
	}
	return s[0]
}
```

Output:

```
Result.Structs : len=3 first=map[]
gconv.Scan     : len=3 first=map[regionId:1]
Result.List()  : len=3 first=map[regionId:1]
```

## Root cause

`Result.Structs` (`database/gdb/gdb_type_result.go`) calls `converter.Structs(...)` directly:

```go
return converter.Structs(r, pointer, gconv.StructsOption{
    SliceOption:  gconv.SliceOption{ContinueOnError: true},
    StructOption: gconv.StructOption{PriorityTag: OrmTagForStruct, ContinueOnError: true},
})
```

In `util/gconv/internal/converter/converter_structs.go`, `Structs` only special-cases whether the
destination element is a **pointer**; otherwise it unconditionally treats it as a **struct**:

```go
reflectElemArray := reflect.MakeSlice(pointerRv.Type().Elem(), len(paramsList), len(paramsList))
itemType         := reflectElemArray.Index(0).Type()   // = map[string]interface{}
itemTypeKind     := itemType.Kind()                     // = reflect.Map
...
if itemTypeKind == reflect.Ptr {
    // pointer element
} else {
    // "Struct element" — taken for reflect.Map too
    ...
    if err = c.Struct(paramsList[i], tempReflectValue, structsOption.StructOption); err != nil {
        return err
    }
    ...
}
```

So each `map` element is fed to `c.Struct`. In `converter_struct.go`, `c.Struct` fetches struct
metadata:

```go
cachedStructInfo := c.internalConverter.GetCachedStructInfo(pointerElemReflectValue.Type(), ...)
if cachedStructInfo == nil {
    return nil   // <-- silent no-op
}
```

and `GetCachedStructInfo` (`util/gconv/internal/structcache/structcache_cached.go`) bails for any
non-struct type:

```go
func (cf *Converter) GetCachedStructInfo(structType reflect.Type, priorityTag string) *CachedStructInfo {
    if structType.Kind() != reflect.Struct {
        return nil
    }
    ...
}
```

Net effect: the pre-allocated slice (correct length) is returned with each map left at its zero
value (`nil`/empty), and `err == nil`.

By contrast, `gconv.Scan` → `converter.Scan` *does* have the map-slice branch
(`converter_scan.go`, `sliceElemKind == reflect.Map` → `MapToMaps`), which is why
`gconv.Scan` / `gconv.Structs` and `Result.List()` all work — only `Result.Structs`
(and therefore `Model.Scan(&[]map...)`) is affected.

## Impact

`Model.Scan(&[]map[string]interface{})` and `Raw(...).Scan(&[]map[string]interface{})` are common
idioms for ad-hoc/untyped queries. The failure is silent (no error, plausible length), so callers
see "the table returned N rows but every field is empty" and easily misattribute it to the driver
or the data.

## Suggested fix (either is fine)

1. **Make it work**: in `converter.Structs`, when `itemTypeKind == reflect.Map`, route through the
   same map-slice path that `converter.Scan` uses (`MapToMaps`) instead of the struct path. This
   makes `Result.Structs(&[]map)` behave like `Result.List()`.
2. **At minimum, fail loudly**: when the destination element is neither struct nor pointer-to-struct
   (e.g. a map), return a clear error rather than silently producing empty elements.

Happy to send a PR for whichever direction maintainers prefer.

sonnemusk:
Fixes #4787

`Result.Structs(&[]map[string]interface{})` returned the right length but every map was empty (no error). `gconv.Scan` / `List()` already worked.

When the destination is a `*[]map[string]…` (interface values), convert via `gconv.Scan(r.List(), pointer)` instead of `converter.Structs`.

Test: `Test_Result_Structs_MapSlice`

```
go test ./database/gdb/ -run Test_Result_Structs_MapSlice
```

### DIFF
--- a/database/gdb/gdb_type_result.go
+++ b/database/gdb/gdb_type_result.go
@@ -9,6 +9,7 @@ package gdb
 import (
 	"database/sql"
 	"math"
+	"reflect"
 
 	"github.com/gogf/gf/v2/container/gvar"
 	"github.com/gogf/gf/v2/encoding/gjson"
@@ -192,6 +193,7 @@ func (r Result) RecordKeyUint(key string) map[uint]Record {
 
 // Structs converts `r` to struct slice.
 // Note that the parameter `pointer` should be type of *[]struct/*[]*struct.
+// It also supports *[]map[string]any / *[]map[string]interface{} (see #4787).
 func (r Result) Structs(pointer any) (err error) {
 	// If the result is empty and the target pointer is not empty, it returns error.
 	if r.IsEmpty() {
@@ -199,6 +201,11 @@ func (r Result) Structs(pointer any) (err error) {
 			return sql.ErrNoRows
 		}
 		return nil
+	}
+	// converter.Structs treats each map row as a struct and leaves []map destinations empty.
+	// Use List()+Scan for map slices so keys/values are preserved.
+	if isMapStringAnySlicePointer(pointer) {
+		return gconv.Scan(r.List(), pointer)
 	}
 	var (
 		sliceOption  = gconv.SliceOption{ContinueOnError: true}
@@ -211,4 +218,21 @@ func (r Result) Structs(pointer any) (err error) {
 		SliceOption:  sliceOption,
 		StructOption: structOption,
 	})
+}
+
+// isMapStringAnySlicePointer reports whether pointer is *[]map[string]T where T is interface{}/any.
+func isMapStringAnySlicePointer(pointer any) bool {
+	rv := reflect.ValueOf(pointer)
+	if rv.Kind() != reflect.Pointer || rv.IsNil() {
+		return false
+	}
+	sliceType := rv.Type().Elem()
+	if sliceType.Kind() != reflect.Slice {
+		return false
+	}
+	elem := sliceType.Elem()
+	if elem.Kind() != reflect.Map {
+		return false
+	}
+	return elem.Key().Kind() == reflect.String && elem.Elem().Kind() == reflect.Interface
 }
--- a/database/gdb/gdb_z_unit_result_structs_map_test.go
+++ b/database/gdb/gdb_z_unit_result_structs_map_test.go
@@ -1,0 +1,39 @@
+// Copyright GoFrame Author(https://goframe.org). All Rights Reserved.
+//
+// This Source Code Form is subject to the terms of the MIT License.
+// If a copy of the MIT was not distributed with this file,
+// You can obtain one at https://github.com/gogf/gf.
+
+package gdb_test
+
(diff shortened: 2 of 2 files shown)


sonnemusk:

PR #4802, opened 2026-07-21
fix(database/gdb): do not map MySQL time_zone into ConfigNode.Timezone

### LINKED ISSUE #4759: gdb: set session variables time_zone cause error
### Go version

go1.24

### GoFrame version

v2

Option Yes

### What did you do?

Image

### What did you see happen?

As mysql golang driver documented, we can set `time_zone='+00:00'` to change connection session time_zone variables.
https://github.com/go-sql-driver/mysql#loc

But when I use it in gf gdb, it cause error: 
```
sql.Open failed for driver \"mysql\" by source \"...?charset=utf8&loc='+00:00'&charset=utf8&loc=UTC&time_zone=%27%2B00%3A00%27&parseTime=True\": unknown time zone ' 00:00'
```

It's seem that the `loc`,`time_zone` is misunderstood in gofr.  all is fine in gorm / raq sql.Open

give time_zone system variables back like https://github.com/go-sql-driver/mysql#system-variables

sonnemusk:
Fixes #4759

Link query `time_zone='%2B00:00'` is a MySQL **session system variable** (see go-sql-driver/mysql). gf parsed Extra with `gstr.Parse` + `gconv.Struct`, which mapped `time_zone` onto `ConfigNode.Timezone`. MySQL open then rewrote that as DSN `loc=...`, e.g.

```
...charset=utf8&loc='+00:00'&charset=utf8&loc=UTC&time_zone=...
```

and `sql.Open` failed with `unknown time zone ' 00:00'`.

Change:

- Drop `time_zone` before Extra → ConfigNode mapping so it stays only in Extra
- Query-escape `loc` when Timezone contains special characters (`/`, `+`, …)

```
go test ./database/gdb/ -run parseConfigNodeLink -count=1
```

### DIFF
--- a/contrib/drivers/mysql/mysql_open.go
+++ b/contrib/drivers/mysql/mysql_open.go
@@ -48,10 +48,13 @@ func configNodeToSource(config *gdb.ConfigNode) string {
 		config.User, config.Pass, config.Protocol, config.Host, portStr, config.Name, config.Charset,
 	)
 	if config.Timezone != "" {
-		if strings.Contains(config.Timezone, "/") {
-			config.Timezone = url.QueryEscape(config.Timezone)
+		// Escape so values like Asia/Shanghai or fixed offsets survive query parsing.
+		// (Previously only names containing "/" were escaped.)
+		loc := config.Timezone
+		if strings.ContainsAny(loc, "/+ :'") {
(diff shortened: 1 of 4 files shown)


sonnemusk:

Issue #4149, opened 2025-02-11
contrib/registry/zookeeper: 服务多实例，会删除之前的注册信息


fengpengyun:
### Go version

go version go1.23.4 windows/amd64

### GoFrame version

2.8.3

### Can this bug be reproduced with the latest release?

Option Yes

### What did you do?

`gsvc.SetRegistry(zookeeper.New(
				[]string{"127.0.0.1:2181"}, zookeeper.WithRootPath("/gogf"),
			))

			s := g.Server("hello.svc")

			s.BindHandler("/", func(r *ghttp.Request) {
				g.Log().Info(r.Context(), `request received`)
				r.Response.Write(`Hello world`)
			})

			s.Run()
`

### What did you see happen?

zookeeper_register.go
`
// ensureName ensure node exists, if not exist, create and set data
func (r *Registry) ensureName(path string, data []byte, flags int32) error {
	exists, stat, err := r.conn.Exists(path)
	if err != nil {
		return gerror.Wrapf(err,
			"Error with check node exist which name is %s",
			path,
		)
	}
	// ephemeral nodes handling after restart
	// fixes a race condition if the server crashes without using CreateProtectedEphemeralSequential()
	if flags&zk.FlagEphemeral == zk.FlagEphemeral {
                // 此处会删除 原有server name对应的node节点
		err = r.conn.Delete(path, stat.Version)
		if err != nil && err != zk.ErrNoNode {
			return gerror.Wrapf(err,
				"Error with delete node which name is %s",
				path,
			)
		}
		exists = false
	}
	if !exists {
		if len(r.opts.user) > 0 && len(r.opts.password) > 0 {
			_, err = r.conn.Create(path, data, flags, zk.DigestACL(zk.PermAll, r.opts.user, r.opts.password))
		} else {
			_, err = r.conn.Create(path, data, flags, zk.WorldACL(zk.PermAll))
		}
		if err != nil {
			return gerror.Wrapf(err,
				"Error with create node which name is %s",
				path,
			)
		}
	}
	return nil
}
`

### What did you expect to see?

保留原有的server name node节点，endpoints对应添加

houseme:
欢迎提PR解决咯，顺便提供一些最小化的实例代码

PR #4808, opened 2026-07-21
fix(contrib/registry/zookeeper): multi-instance registration leaf per endpoint

### LINKED ISSUE #4717: contrib/registry/zk: 多实例服务时，之前的注册信息会被删除
AS TITLE

refer: https://github.com/gogf/gf/issues/4149

sonnemusk:
Fixes #4717 (also #4149)

### Problem
Register used ZK path `.../{serviceName}` as a single ephemeral node and always deleted it before recreate. A second instance of the same service name removed the first instance's registration.

### Fix
1. Leaf path = endpoint-derived name under the service prefix (e.g. `127.0.0.1-9000`)
2. `Search` merges instances with the same service prefix into one entry with combined endpoints
3. `ensureName` only deletes ephemeral nodes when they already exist (avoid nil `stat`)

```
cd contrib/registry/zookeeper && go test -count=1 -run serviceInstanceLeaf .
```

Related prior attempt: #4718 (closed, unmerged).

### DIFF
--- a/contrib/registry/zookeeper/zookeeper_discovery.go
+++ b/contrib/registry/zookeeper/zookeeper_discovery.go
@@ -19,6 +19,9 @@ import (
 
 // Search searches and returns services with specified condition.
 func (r *Registry) Search(_ context.Context, in gsvc.SearchInput) ([]gsvc.Service, error) {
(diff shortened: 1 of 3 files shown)


sonnemusk:

PR #4814, opened 2026-07-28
fix(gsession): atomic write in SetSession to prevent concurrent read race (#4792)

### LINKED ISSUE #4792: [gsession] StorageFile.SetSession 非原子写入导致并发读请求看到空 session

waterWang:
# Fix: Atomic write in `SetSession` to prevent concurrent read race

## Description

`StorageFile.SetSession` uses a non-atomic file write pattern (open with `O_TRUNC` then two separate `Write` calls), creating a race window where concurrent readers see an empty or partially-written file.

### Root Cause

The write path:
1. Opens file with `O_TRUNC` — immediately empties it
2. `file.Write(timestamp)` — writes 8 bytes
3. `file.Write(content)` — writes JSON body

Between steps 1 and 3, any concurrent reader calling `GetSession` sees a file with ≤8 bytes and returns `nil` — the user appears "not logged in".

### Fix

Replace the `O_TRUNC` + two `Write()` calls with atomic write via temp file + `os.Rename()`:

1. Build the complete content (8-byte timestamp + JSON body) in memory
2. Write to a `.tmp` file
3. `os.Rename()` the temp file to the target path

`os.Rename()` is atomic on the same filesystem (POSIX guarantee), so readers always see a complete file.

### Testing

- The existing tests in `os/gsession/gsession_z_unit_storage_file_test.go` cover the session lifecycle
- The fix is a drop-in replacement — same data format, same file path, no new dependencies
- `os.Rename` is used elsewhere in the Go ecosystem for atomic file writes (e.g., `os.WriteFile` itself is atomic on most systems, but we use temp+rename for explicit POSIX atomicity)

Fixes #4792

### DIFF
--- a/os/gsession/gsession_storage_file.go
+++ b/os/gsession/gsession_storage_file.go
@@ -186,20 +186,21 @@ func (s *StorageFile) SetSession(ctx context.Context, sessionId string, sessionD
 			return err
 		}
 	}
-	file, err := gfile.OpenWithFlagPerm(
-		path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, os.ModePerm,
-	)
-	if err != nil {
-		return err
+	// Build content with 8-byte timestamp prefix (same format as before).
+	timestamp := gbinary.EncodeInt64(gtime.TimestampMilli())
+	data := make([]byte, 8+len(content))
+	copy(data[:8], timestamp)
+	copy(data[8:], content)
+
+	// Atomic write: write to temp file first, then rename.
+	// This prevents concurrent readers from seeing an empty/partial file.
+	tmpPath := path + ".tmp"
+	if err := os.WriteFile(tmpPath, data, os.ModePerm); err != nil {
+		return gerror.Wrapf(err, `write data failed to file "%s"`, tmpPath)
 	}
-	defer file.Close()
-	if _, err = file.Write(gbinary.EncodeInt64(gtime.TimestampMilli())); err != nil {
-		err = gerror.Wrapf(err, `write data failed to file "%s"`, path)
-		return err
-	}
-	if _, err = file.Write(content); err != nil {
-		err = gerror.Wrapf(err, `write data failed to file "%s"`, path)
-		return err
+	if err := os.Rename(tmpPath, path); err != nil {
+		os.Remove(tmpPath)
+		return gerror.Wrapf(err, `rename "%s" -> "%s" failed`, tmpPath, path)
 	}
 	return nil
 }
--- a/util/gconv/internal/converter/converter_structs.go
+++ b/util/gconv/internal/converter/converter_structs.go
@@ -115,6 +115,23 @@ func (c *Converter) Structs(params any, pointer any, option ...StructsOption) (e
 			}
 			reflectElemArray.Index(i).Set(tempReflectValue.Addr())
 		}
+	} else if itemTypeKind == reflect.Map {
+		// Map element — use MapToMap instead of Struct.
+		// Struct expects a struct type and silently produces empty maps for non-struct types.
+		for i := 0; i < len(paramsList); i++ {
+			var tempReflectValue reflect.Value
+			if i < pointerRvLength {
+				tempReflectValue = pointerRvElem.Index(i)
+			} else {
+				tempReflectValue = reflect.New(itemType).Elem()
+			}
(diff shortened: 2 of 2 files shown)


lingcoder:
Thanks for looking at #4792. Two blockers, so I'm closing this.

**1. Unrelated change bundled in**

This PR also carries `util/gconv/internal/converter/converter_structs.go`, which is the change from #4815 — the branch appears to have been cut from that one. A session-storage fix shouldn't touch the `gconv` converter.

**2. The temp file name isn't unique**

```go
tmpPath := path + ".tmp"
```

`path` is derived solely from the session id, so two concurrent `SetSession` calls for the *same* session — the exact scenario in the issue — collide on the same temp file and race there instead. The rename is atomic, but the write into `<id>.tmp` is not. `os.CreateTemp(dir, pattern)` followed by `Rename` would make it atomic per-writer.

Worth noting the write also switches from `gfile.OpenWithFlagPerm` to stdlib `os.WriteFile`, while the package uses `gfile` elsewhere.

Happy to review a reworked version. Please include `Test_Issue4792` with concurrent writers and readers under `-race`, asserting that no reader ever observes a truncated file.

PR #4817, opened 2026-07-29
fix(gstr): preserve repeated query parameters in Parse (fix #4751)

### LINKED ISSUE #4751: Bug: `Request.Parse` only keeps the last repeated query value for `[]string`
### Go version

go1.24.0 linux/amd64

### GoFrame version

v2.10.0

Option Yes

### What did you do?

我写了一个最小复现项目，分别用两种请求类型测试 GoFrame 对 GET query 的绑定行为：

1. protobuf 生成的请求结构

```proto
message ListNamesReq {
  repeated string names = 1;
}
```

2. 普通 Go struct

```go
type plainListNamesReq struct {
    Names []string `json:"names"`
}
```

这两个接口内部都使用了 `r.Parse(&req)` 来做参数绑定。

复现步骤如下：

1. 生成 protobuf 代码：

```bash
go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.36.6
PATH="$(go env GOPATH)/bin:$PATH" protoc --go_out=paths=source_relative:. proto/repro.proto
```

2. 启动服务：

```bash
go run .
```

3. 请求 protobuf 路由：

```bash
curl 'http://127.0.0.1:8000/names?names=a&names=b'
```

4. 请求普通 struct 路由：

```bash
curl 'http://127.0.0.1:8000/names-plain?names=a&names=b'
```

### What did you see happen?

实际返回中，`r.Parse(&req)` 绑定后的结果只保留了最后一个值 `b`。

protobuf 路由返回：

```json
{"getQueryValue":"b","parsedNames":["b"],"rawQueryNames":["a","b"],"url":"/names?names=a&names=b"}
```

普通 struct 路由返回：

```json
{"getQueryValue":"b","parsedNames":["b"],"rawQueryNames":["a","b"],"url":"/names-plain?names=a&names=b"}
```

这里可以看到：

- 原始 query `r.Request.URL.Query()["names"]` 仍然是 `[]string{"a", "b"}`
- 但是 `r.Parse(&req)` 绑定到 `[]string` 后只剩下 `[]string{"b"}`

这看起来像是 GoFrame 请求绑定阶段把重复 query 参数覆盖成了最后一个值。

因为目标字段是 `[]string`，并且原始 query 中 `names` 有两个值，所以我期望绑定结果保留全部值：

```json
{"parsedNames":["a","b"]}
```

waterWang:
fix(gstr): preserve repeated query parameters in Parse (fix #4751)

## What

Fix `gstr.Parse` to preserve repeated query parameters when a plain key appears multiple times. Previously, `?name=a&name=b` would only keep the last value (`"b"`), now it correctly produces `["a", "b"]`.

## Why

When a GoFrame request handler receives a URL like `?names=a&names=b` and the target struct field is `[]string`, `r.Parse(&req)` should bind both values. The `gstr.Parse` function's `build` method was overwriting the existing value instead of collecting all values into a slice.

## How

Changed the `build` function's `length == 1` branch to check if the key already exists in the result map. If it does, both the existing and new values are converted to a `[]any` slice and appended. This is consistent with the existing `[]` notation behavior (`v[]=a&v[]=b` → `[a b]`), but without requiring explicit slice notation.

## Documentation updates

The `Parse` function's doc comment is updated to reflect the new behavior:
- `v=m&v=n -> map[v:[m n]]` (was `map[v:n]`)

## Testing

Before:
```
?names=a&names=b → Parse returns {"names": "b"} → bind to []string → ["b"]
```

After:
```
?names=a&names=b → Parse returns {"names": ["a", "b"]} → bind to []string → ["a", "b"]
```

Fixes #4751

### DIFF
--- a/text/gstr/gstr_parse.go
+++ b/text/gstr/gstr_parse.go
@@ -105,7 +105,19 @@ func build(result map[string]any, keys []string, value any) error {
 		key    = strings.Trim(keys[0], "'\"")
 	)
 	if length == 1 {
-		result[key] = value
+		if existing, ok := result[key]; ok {
+			// Key already exists with a plain key (no slice notation).
+			// Convert both values to a slice to preserve all values.
+			// This handles repeated query parameters like ?name=a&name=b.
+			switch v := existing.(type) {
+			case []any:
+				result[key] = append(v, value)
+			default:
+				result[key] = []any{existing, value}
+			}
+		} else {
+			result[key] = value
+		}
 		return nil
 	}
 


lingcoder:
Thanks for picking up #4751 — the underlying complaint is legitimate, `?a=1&a=2` really is multi-valued in HTTP. But this changes the documented contract of `gstr.Parse`, so I'm closing it.

`Parse`'s doc comment defines the notation explicitly:

```
v1=m&v2=n           -> map[v1:m v2:n]
v[]=m&v[]=n         -> map[v:[m n]]
```

A repeated plain key is specified as last-wins; `v[]` is what produces a slice. Turning `v=m&v=n` into `[]any{m, n}` silently changes the return type for every existing caller of `gstr.Parse` — code doing `result["v"].(string)` starts panicking as soon as a client repeats a parameter.

It also interacts badly with the existing conflict handling: `v[a]=n&v=m` would now produce `[]any{map[a:n], "m"}`, a slice mixing a map and a string, where conflicting notations are otherwise meant to be an error.

I think the fix belongs one layer up, in `ghttp`'s parameter binding, where a `[]string` target can consume `url.Values` multi-values without redefining `gstr.Parse` for every caller.

Also: `gstr_parse.go` loses its trailing newline here, and CI has 3 failing jobs.

PR #4821, opened 2026-07-31
fix(contrib/registry/nacos): use Option pattern to prevent data race in SetDefaultMetadata (#4649) [fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT]

waterWang:
## Fix

Use Option pattern to prevent data race in `SetDefaultMetadata` and other setter methods.

### Problem

When `Register` and `SetDefaultMetadata` are called concurrently, Go's race detector reports a data race:

- `SetDefaultMetadata(metadata)` writes to `reg.defaultMetadata` without synchronization
- `Register()` reads `reg.defaultMetadata` without synchronization

The same race exists for `SetClusterName`, `SetGroupName`, and `SetDefaultEndpoint`.

### Solution

Added an `Option` pattern that configures the `Registry` at construction time, which is the idiomatic Go approach for initialization configuration:

- `WithClusterName(name string) Option`
- `WithGroupName(name string) Option`
- `WithDefaultEndpoint(endpoint string) Option`
- `WithDefaultMetadata(metadata map[string]string) Option`

The old `Set*` methods are retained but marked as deprecated, pointing users to the Option alternatives.

### Changes

- `contrib/registry/nacos/nacos.go`:
  - Added `Option` type
  - Added `WithClusterName`, `WithGroupName`, `WithDefaultEndpoint`, `WithDefaultMetadata` Option functions
  - Modified `NewWithClient` to accept `opts ...Option`
  - Modified `NewWithConfig` to accept `opts ...Option` and pass them through
  - Marked `SetClusterName`, `SetGroupName`, `SetDefaultEndpoint`, `SetDefaultMetadata` as deprecated

### Backward Compatibility

- All existing callers continue to work unchanged
- `NewWithClient(client)` with no options works the same as before
- `NewWithConfig(ctx, config)` with no options works the same as before
- `SetClusterName` and `SetGroupName` remain available (deprecated) for legacy callers

### DIFF
--- a/contrib/registry/nacos/nacos.go
+++ b/contrib/registry/nacos/nacos.go
@@ -32,6 +32,41 @@ var (
 	_ gsvc.Registry = &Registry{}
 )
 
+// Option is the option for creating Registry.
+type Option func(*Registry)
+
+// WithClusterName sets the cluster name for the registry.
+// The default is 'DEFAULT'.
+func WithClusterName(name string) Option {
+	return func(r *Registry) {
+		r.clusterName = name
+	}
+}
+
+// WithGroupName sets the group name for the registry.
+// The default is 'DEFAULT_GROUP'.
+func WithGroupName(name string) Option {
+	return func(r *Registry) {
+		r.groupName = name
+	}
+}
+
+// WithDefaultEndpoint sets the default endpoint for service registration.
+// It overrides the service endpoints when registering if it's not empty.
+func WithDefaultEndpoint(endpoint string) Option {
+	return func(r *Registry) {
+		r.defaultEndpoint = endpoint
+	}
+}
+
+// WithDefaultMetadata sets the default metadata for service registration.
+// It will be merged with service's original metadata when registering.
+func WithDefaultMetadata(metadata map[string]string) Option {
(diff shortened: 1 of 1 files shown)


lingcoder:
Thanks for the patch. Closing — this doesn't remove the data race reported in #4649.

The race is that `SetDefaultMetadata` / `SetClusterName` / `SetGroupName` / `SetDefaultEndpoint` write `Registry` fields while concurrent `Register`/`Watch` goroutines read them. This PR adds functional options as an alternative construction path but leaves all four setters in place, annotated `Deprecated`. Existing callers race exactly as before — a deprecation comment isn't enforced by the compiler or by `-race`.

So functionally this is a new public API (`Option`, `WithClusterName`, …) rather than a fix. The option pattern may well be worth having on its own merits, but it should be proposed as such, and the race still needs closing — either by guarding the fields or by removing the setters in a breaking release.

Also: `nacos.go` loses its trailing newline, which `gofmt -l` flags, and CI has 2 failing jobs.

A reworked version with `Test_Issue4649` exercising the setters against concurrent `Register` under `-race` would be welcome.

PR #4822, opened 2026-07-31
fix(gdb/mysql): always URL-encode timezone value in DSN to prevent '+' from being interpreted as space (#4759) [fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT]

### LINKED ISSUE #4759: gdb: set session variables time_zone cause error

waterWang:
fix #4759

## Problem

When setting a MySQL timezone value containing '+' (e.g., `'+00:00'`), the `configNodeToSource` function only URL-encodes the timezone if it contains '/'. The '+' character is not encoded, causing the Go MySQL driver to interpret it as a space in the query string, resulting in:

```
unknown time zone ' 00:00'
```

## Fix

Always URL-encode the timezone value with `url.QueryEscape` instead of only when it contains '/'. This ensures special characters like '+', ':', and quotes are correctly encoded.

```diff
- if strings.Contains(config.Timezone, "/") {
-     config.Timezone = url.QueryEscape(config.Timezone)
- }
+ config.Timezone = url.QueryEscape(config.Timezone)
```

## Verification

- `url.QueryEscape(UTC)` → `UTC` (no change for safe values)
- `url.QueryEscape(Asia/Shanghai)` → `Asia%2FShanghai` (same as before)
- `url.QueryEscape(+00:00)` → `%27%2B00%3A00%27` (now correctly encoded)
- The `strings` import is no longer needed and has been removed.

### DIFF
--- a/contrib/drivers/mysql/mysql_open.go
+++ b/contrib/drivers/mysql/mysql_open.go
@@ -10,7 +10,6 @@ import (
 	"database/sql"
 	"fmt"
 	"net/url"
-	"strings"
 
 	"github.com/gogf/gf/v2/database/gdb"
 	"github.com/gogf/gf/v2/errors/gcode"
@@ -48,9 +47,7 @@ func configNodeToSource(config *gdb.ConfigNode) string {
 		config.User, config.Pass, config.Protocol, config.Host, portStr, config.Name, config.Charset,
 	)
 	if config.Timezone != "" {
-		if strings.Contains(config.Timezone, "/") {
-			config.Timezone = url.QueryEscape(config.Timezone)
-		}
+		config.Timezone = url.QueryEscape(config.Timezone)
 		source = fmt.Sprintf("%s&loc=%s", source, config.Timezone)
 	}
 	if config.Extra != "" {


lingcoder (changes requested):
Thanks for digging into #4759. The encoding observation is correct, but I don't think this change actually fixes the reported problem.

The driver (go-sql-driver/mysql v1.7.1, `dsn.go:468`) handles `loc` like this:

```go
case "loc":
    if value, err = url.QueryUnescape(value); err != nil {
        return
    }
    cfg.Loc, err = time.LoadLocation(value)
    if err != nil {
        return
    }
```

`time.LoadLocation` only accepts IANA names (`Asia/Shanghai`, `UTC`, `Local`) — it does not accept numeric offsets. I ran the DSNs through `mysql.ParseDSN` directly:

| DSN | result |
|---|---|
| `...&loc='+00:00'&...&loc=UTC&...` (as in #4759) | ERR `unknown time zone ' 00:00'` |
| same, with this PR's encoding | ERR `unknown time zone '+00:00'` |
| `loc=+00:00` (master) | ERR `unknown time zone  00:00` |
| `loc=%2B00%3A00` (this PR) | ERR `unknown time zone +00:00` |
| `loc=UTC` only | OK |

So after this PR the connection still fails — only the error text changes. A user setting `timezone: "+00:00"` still cannot connect.

The root cause looks different. `loc` and `time_zone` are two different things:

- `loc` is the **Go-side** location used to parse `DATETIME` values, and needs an IANA name.
- `time_zone` is a **MySQL session variable**; the driver does not treat it as a known key and passes it through verbatim (`Params=map[time_zone:'+00:00']`), which is why `'+00:00'` is valid there.

`configNodeToSource` puts the `timezone` config into `loc`, so an offset value can never work. Two things worth considering:

1. Reject or convert offset-style `timezone` values instead of emitting an unusable `loc`, and/or document that `timezone` must be an IANA name (`time_zone` belongs in `extra`).
2. The value in #4759 also carries single quotes (`'+00:00'`), which `loc` won't accept in any case.

Note the DSN in #4759 has `loc` twice (one from `config.Timezone`, one from `extra`). That is not itself the failure though — `parseDSNParams` returns as soon as the first `loc` fails to parse, so the later `loc=UTC` is never reached.

If the encoding cleanup is kept, it should come with a test (`contrib/drivers/mysql/mysql_z_unit_issue_test.go`, `Test_Issue4759`) asserting the connection actually works, not just that the string is encoded. Happy to review a reworked version.

lingcoder:
Closing this per the review above: after the change the connection still fails, only the error text changes, because `loc` requires an IANA name and `+00:00` isn't one.

The root cause is elsewhere — `configNodeToSource` maps the `timezone` config onto the driver's `loc` parameter, while the value in #4759 is meant for the MySQL `time_zone` session variable. Those are two different things.

Happy to review a reworked version along those lines, with `Test_Issue4759` in `contrib/drivers/mysql/mysql_z_unit_issue_test.go` asserting that the connection actually succeeds, rather than just that the string is encoded.

PR #4829, opened 2026-08-04
fix(gdb): Model.Partition() now emits PARTITION clause in SELECT queries

### LINKED ISSUE #4826: database/gdb: Model.Partition() is a no-op, partition names never reach the generated SQL

waterWang:
## Description

`Model.Partition()` set the `m.partition` field but it was never read during SQL generation. The `PARTITION` clause was absent from every SELECT query, so `Partition()` was effectively a no-op — every call returned the whole table regardless of the partition name.

### Root cause

`database/gdb/gdb_model.go`:
```go
func (m *Model) Partition(partitions ...string) *Model {
    model := m.getModel()
    model.partition = gstr.Join(partitions, ",")
    return model
}
```

The `partition` field had exactly two references in the whole repository — the declaration and the setter. It was never read by `getFormattedSqlAndArgs()` in `gdb_model_select.go`, which generates the SQL string directly from `m.tables`.

### Fix

1. **Add `GetPartitionClause(partitions string) string` to the `DB` interface** — provides a driver-extensible way to emit the `PARTITION` clause.
2. **Default implementation in `Core`** — returns ` PARTITION (partitions)` for MySQL/MariaDB/TiDB.
3. **PostgreSQL and GaussDB drivers override** — return empty string (these databases don't support MySQL-style `PARTITION (p)` clause in SELECT; they use declarative partitioning with inheritance-based pruning).
4. **`getFormattedSqlAndArgs` now uses `getTableWithPartition()`** — appends the partition clause after the table name when `m.partition` is set.
5. **Test expected values updated** — `Partition("p3")` returns 1 row (was 5, which was the whole-table count).

### Before

```sql
SELECT * FROM `t`
SELECT * FROM `t` WHERE ...  -- Partition("p3") had no effect
```

### After (MySQL/MariaDB)

```sql
SELECT * FROM `t` PARTITION (p3)
SELECT * FROM `t` PARTITION (p3) WHERE ...
```

### After (PostgreSQL/GaussDB — unchanged, no-op with warning in docs)

No change — `PARTITION` clause is not supported on these drivers.

Fixes #4826

### DIFF
--- a/contrib/drivers/gaussdb/gaussdb_dialect.go
+++ b/contrib/drivers/gaussdb/gaussdb_dialect.go
@@ -24,3 +24,10 @@ func (d *Driver) GetBoolLiteral(v bool) string {
 func (d *Driver) GetLockSharedClause() string {
 	return gdb.LockForShare
 }
+
+// GetPartitionClause returns an empty string for GaussDB. GaussDB, being
+// PostgreSQL-compatible, does not support the MySQL-style PARTITION (p)
+// clause on SELECT statements.
+func (d *Driver) GetPartitionClause(partitions string) string {
+	return ""
+}
--- a/contrib/drivers/mariadb/mariadb_unit_init_test.go
+++ b/contrib/drivers/mariadb/mariadb_unit_init_test.go
@@ -189,7 +189,7 @@ func Test_PartitionTable(t *testing.T) {
 		data, err = db3.Ctx(ctx).Model("dbx_order").Partition("p3").All()
 		t.AssertNil(err)
 		dataLen = len(data)
-		t.Assert(dataLen, 5)
+		t.Assert(dataLen, 1)
 	})
 }
 
--- a/contrib/drivers/mysql/mysql_z_unit_init_test.go
+++ b/contrib/drivers/mysql/mysql_z_unit_init_test.go
@@ -187,7 +187,7 @@ func Test_PartitionTable(t *testing.T) {
 		data, err = db3.Ctx(ctx).Model("dbx_order").Partition("p3").All()
 		t.AssertNil(err)
 		dataLen = len(data)
-		t.Assert(dataLen, 5)
+		t.Assert(dataLen, 1)
 	})
 }
 func createShopDBTable() {
--- a/contrib/drivers/pgsql/pgsql_dialect.go
+++ b/contrib/drivers/pgsql/pgsql_dialect.go
@@ -25,3 +25,10 @@ func (d *Driver) GetBoolLiteral(v bool) string {
 func (d *Driver) GetLockSharedClause() string {
 	return gdb.LockForShare
 }
+
+// GetPartitionClause returns an empty string for PostgreSQL. PostgreSQL does
+// not support the MySQL-style PARTITION (p) clause on SELECT statements;
+// it uses declarative partitioning with inheritance-based partition pruning.
+func (d *Driver) GetPartitionClause(partitions string) string {
+	return ""
+}
--- a/database/gdb/gdb.go
+++ b/database/gdb/gdb.go
@@ -356,6 +356,12 @@ type DB interface {
 	// Drivers that don't support MySQL's legacy "LOCK IN SHARE MODE" override
 	// to return their dialect equivalent (e.g. "FOR SHARE" on PostgreSQL).
 	GetLockSharedClause() string
+
+	// GetPartitionClause returns the SQL PARTITION clause appended to the
+	// table name in SELECT queries. Returns an empty string when the driver
+	// does not support the PARTITION clause (e.g. PostgreSQL).
+	// MySQL/MariaDB syntax: " PARTITION (p0,p1)"
+	GetPartitionClause(partitions string) string
 }
 
 // TX defines the interfaces for ORM transaction operations.
--- a/database/gdb/gdb_core_underlying.go
+++ b/database/gdb/gdb_core_underlying.go
@@ -530,6 +530,17 @@ func (c *Core) GetLockSharedClause() string {
 	return LockInShareMode
 }
 
+// GetPartitionClause returns the SQL PARTITION clause for the given partition
+// names. MySQL/MariaDB syntax: " PARTITION (p0,p1)". Drivers that do not
+// support the PARTITION clause (e.g. PostgreSQL, GaussDB) override this to
+// return an empty string.
+func (c *Core) GetPartitionClause(partitions string) string {
+	if partitions == "" {
+		return ""
+	}
+	return " PARTITION (" + partitions + ")"
+}
+
 func (c *Core) columnValueToLocalValue(ctx context.Context, value any, columnType *sql.ColumnType) (any, error) {
 	var scanType = columnType.ScanType()
 	if scanType != nil {
--- a/database/gdb/gdb_model_select.go
+++ b/database/gdb/gdb_model_select.go
@@ -746,6 +746,15 @@ func (m *Model) doGetAllBySql(
 func (m *Model) getFormattedSqlAndArgs(
 	ctx context.Context, selectType SelectType, limit1 bool,
 ) (sqlWithHolder string, holderArgs []any) {
+	// getTableWithPartition returns m.tables with the driver-specific
+	// PARTITION clause appended when m.partition is set.
+	getTableWithPartition := func() string {
+		table := m.tables
+		if m.partition != "" {
+			table += m.db.GetPartitionClause(m.partition)
+		}
+		return table
+	}
 	switch selectType {
 	case SelectTypeCount:
 		queryFields := "COUNT(1)"
@@ -764,7 +773,7 @@ func (m *Model) getFormattedSqlAndArgs(
 			return sqlWithHolder, conditionArgs
 		}
 		conditionWhere, conditionExtra, conditionArgs := m.formatCondition(ctx, false, true)
-		sqlWithHolder = fmt.Sprintf("SELECT %s FROM %s%s", queryFields, m.tables, conditionWhere+conditionExtra)
+		sqlWithHolder = fmt.Sprintf("SELECT %s FROM %s%s", queryFields, getTableWithPartition(), conditionWhere+conditionExtra)
 		if len(m.groupBy) > 0 {
 			sqlWithHolder = fmt.Sprintf("SELECT COUNT(1) FROM (%s) count_alias", sqlWithHolder)
 		}
@@ -785,7 +794,7 @@ func (m *Model) getFormattedSqlAndArgs(
 		// DISTINCT t.user_id uid
 		sqlWithHolder = fmt.Sprintf(
 			"SELECT %s%s FROM %s%s",
-			m.distinct, m.getFieldsFiltered(), m.tables, conditionWhere+conditionExtra,
+			m.distinct, m.getFieldsFiltered(), getTableWithPartition(), conditionWhere+conditionExtra,
 		)
 		return sqlWithHolder, conditionArgs
 	}


lingcoder (changes requested):
The direction and the layering here are right — this is a much better take than #4827.

Putting `GetPartitionClause` on the `DB` interface, with a MySQL default on `Core` and per-driver overrides, mirrors the existing `GetLockSharedClause` pattern, and `DriverWrapperDB` embeds `DB` so the overrides dispatch correctly. It also no longer rewrites `m.tables`, so it doesn't disturb the field-filtering branch in `gdb_model_select.go` that keys off `gstr.Contains(m.tables, " ")`, nor `guessPrimaryTableName` in `gdb_model_cache.go`. Both SELECT paths (count and regular) are covered.

One thing needs fixing first, plus a few trade-offs I'd like to confirm.

---

## Needs fixing first: partition names are concatenated into SQL unvalidated

`Model.Partition` does nothing to its input:

```go
func (m *Model) Partition(partitions ...string) *Model {
	model := m.getModel()
	model.partition = gstr.Join(partitions, ",")
	return model
}
```

And this PR's default implementation concatenates raw:

```go
func (c *Core) GetPartitionClause(partitions string) string {
	if partitions == "" {
		return ""
	}
	return " PARTITION (" + partitions + ")"
}
```

There is no validation or quoting anywhere between the argument and the final SQL string. Partition names are identifiers, so they cannot be parameterized with placeholders and have to be handled explicitly — gf does quote identifiers elsewhere (for example JOIN aliases go through `QuoteWord` in `gdb_model_join.go`), so this is also inconsistent with the surrounding code.

What matters here: **`m.partition` was never read before this PR, so the risk did not exist; this change is what wires it into SQL generation, which makes the injection surface newly introduced.** Selecting a partition dynamically by tenant or by date is a natural usage, and that value can easily originate from a request.

Two ways to address it:

- Validate in `Partition()` — restrict partition names to `[A-Za-z0-9_]+` and reject anything else
- Or `QuoteWord` each partition name inside `GetPartitionClause` before joining

---

## Trade-offs worth confirming

**1. SELECT only**

Per the MySQL manual, [explicit partition selection](https://dev.mysql.com/doc/refman/8.0/en/partitioning-selection.html) is supported for `SELECT` / `DELETE` / `INSERT` / `REPLACE` / `UPDATE` / `LOAD DATA` / `LOAD XML`. This PR only touches `getFormattedSqlAndArgs`, so `Partition()` remains a silent no-op on writes.

The title does say "in SELECT queries", so the scope is stated honestly — but #4826 describes `Partition()` as broken in general. Is the plan to cover writes in a follow-up, or to keep the scope here? If the latter, it's worth documenting the limitation in the `Partition()` doc comment.

**2. Drivers without an override will emit MySQL-specific syntax**

Only pgsql and gaussdb override to return an empty string. clickhouse / dm / mssql / oracle / sqlite / sqlitecgo inherit the `Core` default and would emit `PARTITION (...)`, producing a syntax error.

`GetLockSharedClause` has exactly the same gap, so this is a property of the existing pattern rather than something this PR introduces. Still, this PR turns what is currently a silent no-op on those six drivers into a runtime error, which is a behaviour change worth agreeing on up front.

**3. Should unsupported drivers error or silently ignore?**

#4826 carries the `discuss` label precisely because this is unsettled. This PR picks silent-empty-string, meaning `Partition("p0")` on PostgreSQL returns the full table with no indication that the partition filter was dropped. That trade-off needs a maintainer decision.

---

## Tests

Changing the assertion from `dataLen == 5` to `1` is correct. Working it out from the fixture: the partition key is `YEAR(sales_date)` and `insertShopDBData` inserts 5 rows for years 2021 through 2025, so p1=0, p2=0, **p3=1**, p4=4. After the fix `Partition("p3")` should indeed return 1 row, while `Partition("p3","p4")` stays at 5 — leaving the first assertion untouched is correct, not an oversight.

That also confirms the original `Partition("p3") == 5` had locked in the no-op behaviour: selecting one partition or two both returned the whole table.

That said, the repo convention is for an issue fix to add `Test_Issue4826` rather than only adjust existing assertions. The injection point above needs a case as well. Please add both.

lingcoder:
Closing this one as well. See #4831 for the fuller explanation.

To be fair on the technical side: the layering in this PR is sound, and I said so in the review above. The blocker was the unvalidated partition name concatenated straight into SQL — an injection surface this change is what introduces, since `m.partition` was never read before.

But the code is not why this is being closed. #4826 has now received three PRs — #4827, this one, and #4831 — and #4831 was opened while this one was still under review, proposing a different implementation of the same fix. Reviewing competing PRs for a single issue in parallel is not something maintainers can absorb, and it is not clear which one you actually intend to land.

**Please do not use automated tooling to submit unverified PRs in bulk.**

If you want to pursue this properly, open a single PR that:

- addresses the identifier-quoting issue raised in the review above
- includes `Test_Issue4826`, plus a case covering the quoting
- has green CI
- states whether the scope is SELECT only, and documents that limitation in the `Partition()` doc comment if so

And note that #4826 carries the `discuss` label: whether drivers without partition-selection support should error or silently ignore the clause is unsettled. That decision needs to come before a fourth implementation attempt.

PR #4836, opened 2026-08-06
fix(contrib/drivers/oracle): fix GBK decode and quoted identifiers

yuyinw:
### Why

The Oracle driver had two independent issues:

1. `github.com/sijms/go-ora/v2 v2.7.10` can panic while decoding truncated GBK input. In charset case `0x354`, the old guard allowed a trailing one-byte slice to reach `binary.BigEndian.Uint16(input[index:])`, which requires two bytes.
2. `Driver.DoFilter` removed every double quote from SQL before parsing. Oracle treats double-quoted identifiers as case-sensitive, so raw SQL targeting mixed-case table or column names was sent as unquoted identifiers and Oracle uppercased them.

### What changed

- Upgrade `github.com/sijms/go-ora/v2` to `v2.7.18`, the smallest upstream release that includes the GBK decode bounds-check fix.
- Preserve explicit double quotes in raw SQL filtering so users can address case-sensitive Oracle identifiers.
- Return empty quote chars for generated Oracle SQL so normal framework table/model operations keep using Oracle's unquoted identifier behavior.
- Add focused tests for the GBK trailing lead byte regression and the Oracle SQL filter behavior.
- Record both feedback fixes in OpenSpec.

### Testing

- `go test -run 'TestDriver(DoFilterPreservesQuotedIdentifiers|GetCharsDoesNotAddImplicitQuotes)$|^TestStringConverterDecodeGBKTrailingLeadByte$' ./...`
- `go test -run '^$' ./...`


gqcn:
The go-ora bump plus keeping quoted Oracle identifiers looks well scoped, and the gtest coverage for GBK decode and `DoFilter` is helpful. One merge concern:

- **Suggested priority** `openspec/changes/fix-go-ora-string-converter-panic/`: merging this directory would leave an unarchived OpenSpec change on `master`. Community PRs do not need OpenSpec files in the tree; consider dropping that directory from the PR (keep the driver fix and tests).

PR #4848, opened 2026-08-17
fix(util/gconv): unwrap reflect.Value and support pointer element slices

hailaz:
Extracted from #4716: these `gconv` bugs already exist on master and are independent of the Scan-into-basic-type feature.

## Problem

On master:

```go
gconv.Int64(reflect.ValueOf(42))          // 0
gconv.Bool(reflect.ValueOf(false))        // true
gconv.String(reflect.ValueOf(42))         // "<int Value>"

vs := gvar.Vars{gvar.New(1), gvar.New(2)}
var ids []*int64
vs.Scan(&ids)                             // panic: SetInt on ptr Value
```

`Bool`/`Int64`/`Uint64`/`Float*` treat `reflect.Value` as the converted value. `String()` hits `reflect.Value.String()` and returns the reflection type name. Slice `Scan` into `[]*T` calls `SetInt`/`SetString` on a pointer value.

## Fix

- Unwrap interfaceable `reflect.Value` inputs in the basic converters
- Allocate pointer slice elements before assigning scanned values

## Test plan

- [x] `go test -count=1 ./util/gconv/`
- [x] `TestScanReflectValueInput`
- [x] `TestScanPointerElementSlice`

Related: #4716 (feature PR should rebase onto this after merge)

### DIFF
--- a/util/gconv/gconv_z_unit_scan_basic_types_test.go
+++ b/util/gconv/gconv_z_unit_scan_basic_types_test.go
@@ -7,6 +7,7 @@
 package gconv_test
 
 import (
+	"reflect"
 	"testing"
 	"time"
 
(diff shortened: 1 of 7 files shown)


lingcoder (approved):
LGTM

PR #4854, opened 2026-08-18
fix(os/grpool): stop Test_Limit4 asserting a sum that has a race window

### LINKED ISSUE #4853: os/grpool: Test_Limit4 is flaky, its assertion has a race window by design
### Go version

go1.23 / go1.24 / go1.25 (all matrix entries, on CI runners)

### GoFrame version

master (9edefe991)

Yes

### What did you do?

Ran `Test_Limit4` on CI. It is a flaky test rather than a library defect — the assertion at `os/grpool/grpool_z_unit_test.go:155` is not an invariant:

```go
for i := 0; i < size; i++ {          // size = 1000, pool limit = 100
    pool.Add(ctx, func(ctx context.Context) {
        array.Append(1)
        time.Sleep(2 * time.Second)
    })
}
t.Assert(waitUntil(...), true)
t.Assert(pool.Jobs()+array.Len(), size)   // line 155
```

The same assertion appears three times in this test: lines 155, 162 and 174.

To reproduce, run it on a loaded machine (an idle one will not do it — I ran it 11 times locally, with and without `-race`, all green).

### What did you see happen?

Intermittent failures in CI, with a varying shortfall:

```
[ASSERT] EXPECT 998 == 1000
--- FAIL: Test_Limit4 (2.00s)
FAIL	github.com/gogf/gf/v2/os/grpool

[ASSERT] EXPECT 962 == 1000
--- FAIL: Test_Limit4 (0.30s)
FAIL	github.com/gogf/gf/v2/os/grpool
```

Because `fail-fast` is enabled, this single job cancels the rest of the matrix, so unrelated pull requests appear broken.

The test should pass consistently.

A job leaves the queue before it runs `array.Append(1)`. During that window it is counted in neither `Jobs()` nor `array.Len()`, so the sum can only be bounded, never exact:

```
size - limit  <=  Jobs() + array.Len()  <=  size
```

With a limit of 100 there can be up to 100 jobs in that window at once, which matches the observed shortfalls of 2 and 38. The two values are also read separately, so even without the window the sum is not a consistent snapshot.

lingcoder:
## Summary

`Test_Limit4` fails intermittently on loaded CI runners, and because `fail-fast` is on, it cancels the rest of the matrix and makes unrelated pull requests look broken.

```
[ASSERT] EXPECT 998 == 1000
[ASSERT] EXPECT 962 == 1000
--- FAIL: Test_Limit4
```

## Why it fails

```go
pool.Add(ctx, func(ctx context.Context) {
    array.Append(1)          // a job is in neither place until it reaches here
    time.Sleep(2 * time.Second)
})
...
t.Assert(pool.Jobs()+array.Len(), size)
```

A job leaves the queue before it appends to the array. In that window it is counted in neither `Jobs()` nor `array.Len()`, so while the pool is running the sum can only be bounded:

```
size - limit  <=  Jobs() + array.Len()  <=  size
```

With a limit of 100, up to 100 jobs can sit in that window at once — matching the observed shortfalls of 2 and 38. The two values are also read separately, so the sum is not a consistent snapshot either.

It is load dependent, which is why it only shows up on CI: 11 local runs with and without `-race` were all green.

## Changes

The three assertions taken while jobs are running now bound the sum through a small helper. The fourth one, after `pool.Close()` when `Size()` is 0 and no worker can be in the window, keeps the exact equality — there the invariant does hold.

closes #4853

### DIFF
--- a/os/grpool/grpool_z_unit_test.go
+++ b/os/grpool/grpool_z_unit_test.go
@@ -18,6 +18,17 @@ import (
 	"github.com/gogf/gf/v2/test/gtest"
 )
 
+// assertNoJobLost checks that no job went missing, without requiring an exact sum.
+// A job leaves the queue before it appends to the array, so during that window it
+// is counted in neither Jobs() nor the array. inFlightLimit bounds how many jobs
+// can be in that window: the peak worker count the pool has had, not its current
+// Cap(), because jobs left over from before a shrink are still in flight.
+func assertNoJobLost(t *gtest.T, pool *grpool.Pool, array *garray.Array, size, inFlightLimit int) {
+	sum := pool.Jobs() + array.Len()
+	t.AssertLE(sum, size)
+	t.AssertGE(sum, size-inFlightLimit)
+}
+
 func waitUntil(timeout time.Duration, condition func() bool) bool {
 	deadline := time.Now().Add(timeout)
 	for time.Now().Before(deadline) {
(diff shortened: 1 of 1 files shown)


houseme (approved):
LGTM

PR #4874, opened 2026-09-06
feat(contrib/drivers/sqlite): add loong64 support

lijunting2020:
To add support for the loong64 architecture in goframe, we bumped the dependencies in the cmd/gf and contrib/drivers/sqlite modules: modernc.org/sqlite from v1.23.1 to v1.30.0, and modernc.org/libc from v1.22.5 to v1.50.9."


gqcn:
Thanks for working on loong64 support. The dependency bump itself looks like a reasonable direction. One process item:

- **Suggested improvement** title: `feat(cmd/gf contrib/drivers/sqlite): to support the loong64 architecture` is a bit hard to scan. Consider something shorter such as `feat(contrib/drivers/sqlite): add loong64 support`.

lijunting2020:
Thanks! I've updated the PR title to:
feat(contrib/drivers/sqlite): add loong64 support

Just to confirm, do I still need to squash the 11 commits into a single commit?

PR #4882, opened 2026-09-18
fix(database/gdb): map database type names exactly in each driver instead of guessing from substrings

### LINKED ISSUE #4842: database/gdb: type detection matches "int" as a substring and takes point/interval for integers
### 问题

`database/gdb/gdb_core_structure.go:360`，驱动未显式列举的类型会落到 Core 的
自动探测分支，那里用子串匹配猜类型：

```go
case strings.Contains(typeName, "int"):
    if gstr.ContainsI(fieldType, "unsigned") {
        return LocalTypeUint, nil
    }
    return LocalTypeInt, nil
```

凡是类型名里含 `int` 的都被当成整数。PostgreSQL / openGauss 上至少这些不是
整数却会命中：

`point`（p-o-**int**）、`interval`、`tinterval`、`int4range`、`int8range`、
`int4multirange`、`int8multirange`，以及它们的数组形式。

后果是值被转成 0：

```
point    存 (1.5,2.5)   读回 0
interval 存 '2 days'    读回 0
```

数据本身没问题，是读取时的类型判定错了。

### 佐证

- 该分支的匹配顺序是 text/char → float/double/numeric → bool → binary/blob →
  **int** → time → date，所以 `point` 会先命中 `int`，走不到后面。
- 目前各驱动是靠在自己的 `CheckLocalTypeForField` 里**逐个列举**来绕过它的。
  pgsql 那份清单的 13 条 case 全是被报了 bug 之后才补的：
  `bytea` 见 #4231 / #4677（由 #4678 修）、`numeric[]` / `decimal[]` 见 #4457
  （由 #4511 修）。
- #4840 又补进了 `point` / `interval` / range 系列。

### 影响

清单式绕过只能挡住已经被发现的类型：任何未列举、名字里含 `int` 的类型仍会被
静默转成 0，而且每个驱动都要各自补一遍。

### 可能的修法

- 在 Core 里加一份「已知非整数」的排除清单；
- 或把子串匹配收紧（按词边界匹配，或改为精确类型名表）。

任一改法都会影响所有驱动，需要评估 dm / oracle / mssql / clickhouse 等本地
难以验证的驱动，所以先开 issue 记录，不直接改。

### 环境

- master (41baf1e28)
- 实测数据库：openGauss 7.0.0-RC1、PostgreSQL 17

lingcoder:
closes #4842

## Problem

`Core.CheckLocalTypeForField` falls back to keyword matching on the type name, so any name that merely embeds a keyword is misclassified. Every value of such a type is silently converted to 0:

| Database | Type names read back as 0 |
|---|---|
| PostgreSQL / openGauss | `point`, `interval`, `int4range`, `int8range`, `int2vector` and their array forms |
| Oracle | `IntervalDS_DTY`, `IntervalYM_DTY` |
| DM | all 13 `INTERVAL ...` variants |
| ClickHouse | `Point`, `IntervalSecond` … `IntervalYear` |
| SQLite / sqlitecgo | any column declared `point`, `interval`, etc. |

The same matching takes PostgreSQL's `date[]`, `time[]`, `timestamp[]` and `daterange[]` for a single `datetime`, dropping every element but the first.

Reproduction, SQLite, no dependencies:

```sql
CREATE TABLE t(p point, txt varchar(64));
INSERT INTO t VALUES('(1.5,2.5)', '(1.5,2.5)');
SELECT * FROM t;   -- p: 0    txt: (1.5,2.5)
```

## Change

Every driver now maps the type names it can report **exactly**, from the authoritative list of its underlying driver or server, and falls back to the core only for a name outside that list:

| Driver | Source of the type list | Entries |
|---|---|---|
| pgsql | `oid.TypeName` of `lib/pq` | 172 |
| gaussdb | `oid.TypeName` of the openGauss driver | 209 |
| oracle | `TNSType.String()` of `go-ora`, enumerated | 44 |
| clickhouse | `system.data_type_families` of the server | 136 |
| dm | a column declared of every DM8 type, read back | 58 |

Each driver gains a `Test_LocalTypeCoverage` that fails when the driver or server reports a name missing from its map, so the map cannot silently rot as the underlying driver gains types. pgsql and gaussdb also gain the reverse check, so the map cannot keep stale entries.

`pgsql` and `gaussdb` previously kept a second, parallel list of type names in `ConvertValueForLocal`; it now dispatches on the `LocalType` instead, so a type name is classified in one place only.

SQLite has no finite set of type names — a declared type is free text, and by [its own rules](https://www.sqlite.org/datatype3.html#affinity) `point` gets INTEGER affinity because it contains `INT`. Its driver therefore trusts the storage class of the value over the declared type: when the name suggests a number but the driver hands over text, SQLite has already determined the value is not a number.

## Behaviour change

PostgreSQL / openGauss arrays of temporal types (`date[]`, `time[]`, `timestamp[]`, `timestamptz[]`, `timetz[]`, `daterange[]`) now read as `[]string` of their elements rather than a single `datetime` holding only the first element.

## Out of scope, to follow

- The keyword matching in `Core.CheckLocalTypeForField` is untouched; with every driver mapping its own names it is now reachable only from drivers without a map (mysql, mariadb, mssql, tidb, oceanbase) and can be removed once those have theirs.
- `Test_New_Path_With_Colon` in sqlite/sqlitecgo fails on Windows independently of this change (it creates a directory named `test:1`, which Windows forbids).

## Verification

All driver suites run locally against the CI images with `-count=1`: pgsql, gaussdb, oracle, clickhouse, dm, mysql, mariadb, mssql pass; sqlite and sqlitecgo pass except the pre-existing Windows failure above. Every `Test_Issue4842` was confirmed red on master and green with the change; every coverage test was confirmed to name a deliberately removed entry.


houseme (approved):

gqcn:
The exact maps fix the `point` / `interval` misreads, and the `bit(1)` plus SQLite REAL follow-ups look right. One pair of entries still throws the value away the same way the old keyword match did.

`daterange` is `LocalTypeDatetime` in `contrib/drivers/pgsql/pgsql_local_type.go:132` and `contrib/drivers/gaussdb/gaussdb_local_type.go:140`. A value like `[2026-01-01,2026-02-01)` then goes through `gtime.StrToTime`, which keeps the first date and drops the upper bound, so the caller sees `2026-01-01 00:00:00` instead of the range. `tsrange`, `tstzrange`, `int4range` and `numrange` in those same maps are already strings, and `_daterange` was switched to `[]string` on purpose. `reltime` is the same split: `_reltime` is a string slice, but the scalar at `pgsql_local_type.go:187` and `gaussdb_local_type.go:215` is still a datetime. Something like `1 day` does not parse, the core conversion ignores that error, and the original text is gone.

Mapping both scalars to `LocalTypeString` would keep the text. A scalar `SELECT '[2026-01-01,2026-02-01)'::daterange` next to the existing array case in `Test_Issue4842` would lock that in.

The note above `ConvertValueForLocal` in `pgsql_convert.go` and `gaussdb_convert.go` still says `_date`, `_timestamp` and `_timestamptz` are not mapped. They are, and `Test_Issue4842` already expects `[]string`. That note should match the map, otherwise the next edit will "fix" those arrays back. `_json` / `_jsonb` are mapped to `LocalTypeString` rather than a slice, so if that part is still intentional, the note can say so directly.

Leaving `bot-approved` off until `daterange` and `reltime` read back as text.

gqcn:
上面那条是英文，这里用中文再说一遍，结论一样。

类型表这次改对了大半。`point` 和 `interval` 不会再被认错，`bit(1)` 和 SQLite 的 REAL 也没看出问题。剩下 `daterange` 和 `reltime` 这两个标量，查出来以后原文丢了，和以前按关键字猜类型时一样。

PostgreSQL 和 GaussDB 里，`daterange` 还标着 `LocalTypeDatetime`，在 `contrib/drivers/pgsql/pgsql_local_type.go:132` 和 `contrib/drivers/gaussdb/gaussdb_local_type.go:140`。`[2026-01-01,2026-02-01)` 这种区间会交给 `gtime.StrToTime`，只解析出开头那天，结束边界被丢掉，调用方看到的是 `2026-01-01 00:00:00`。同一张表里，`tsrange`、`tstzrange`、`int4range`、`numrange` 已经是字符串，数组 `_daterange` 也改成了 `[]string`，就这两个标量还留在时间类型上。

`reltime` 也是这样。数组 `_reltime` 是字符串切片，标量还是 datetime，见 `pgsql_local_type.go:187` 和 `gaussdb_local_type.go:215`。`1 day` 这种值解析不了，转换层把错误忽略掉，原来的文本就没了。

标量改成 `LocalTypeString`，文本就能留下来。`Test_Issue4842` 里已经有数组用例，旁边补一条 `SELECT '[2026-01-01,2026-02-01)'::daterange`，这个行为就固定住了。

另外有一处注释和代码对不上。`pgsql_convert.go` 和 `gaussdb_convert.go` 里，`ConvertValueForLocal` 上面还写着 `_date`、`_timestamp`、`_timestamptz` 没有映射。表里其实已经有了，`Test_Issue4842` 也在检查 `[]string`。注释不改的话，下次有人按注释改代码，容易把这些数组改回去。`_json` 和 `_jsonb` 映射的是 `LocalTypeString`，不是切片；如果这是故意的，在注释里写清楚就行。

`daterange` 和 `reltime` 能按文本读回来之后，再打 `bot-approved`。

gqcn:
`daterange` / `reltime` now stay text, and the SQLite REAL case is covered. One openGauss mapping still drops the value the same way the old substring match did.

In `contrib/drivers/gaussdb/gaussdb_local_type.go`, `_int1` and `_int16` are arrays (the same `_` prefix as `_int2`) but both are `LocalTypeInt` (lines 59–60). The driver hands the array over as text, like `{1,2,255}`, and `LocalTypeInt` runs that through `gconv.Int`, so the caller gets `0`. `_int2` next to them is already `[]int32`. `int1[]` is a normal tinyint array, so this shows up on user tables.

`int16` (line 156) is not a 16-bit integer either. openGauss defines it as a 16-byte integer; `LocalTypeInt` still truncates it. `LocalTypeBigInt` keeps the decimal text. You can't declare an `int16` column today, but the array form is the one that loses data the same way `point` used to. `_int1` can be `LocalTypeInt32Slice` (those values fit); `_int16` needs `LocalTypeStringSlice` so a 128-bit element is not forced into `int32`. A `SELECT '{1,2}'::int1[]` next to `Test_Issue4842` would lock the array case in.

